PR #1, opened 2016-11-09
First-pass implementation of the load API

cwillisf:
This implementation is untested and not complete, but should be sufficient for discussion.

### DIFF
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -1,7 +1,8 @@
 module.exports = {
+    root: true,
+    extends: ['scratch', 'scratch/es6', 'scratch/node'],
     env: {
         node: false,
         browser: false
-    },
-    extends: ['scratch/es6']
+    }
 };


src/BuiltinHelper.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * @type {BuiltinAssetRecord[]}
+ */
+const DefaultAssets = [
+    {
+        type: AssetType.ImageBitmap,
+        format: DataFormat.PNG,
+        id: 'e5cb3b2aa4e1a9b4c735c3415e507e66',
+        data: require('binary!./builtins/e5cb3b2aa4e1a9b4c735c3415e507e66.png') // eslint-disable-line global-require
```
cwillisf: This causes Webpack to insert the contents of the file as a string literal. That way we can guarantee access to these few assets even if (for example) Scratch is loaded from a web page and then the network connection is immediately terminated.

src/BuiltinHelper.js
```diff
@@ -0,0 +1,98 @@
...
+     * @return {Promise.<Asset>} A promise for the contents of the asset.
+     */
+    load (assetType, assetId) {
+        if (this.assets.hasOwnProperty(assetType.name)) {
+            const typeBucket = this.assets[assetType.name];
+            if (typeBucket.hasOwnProperty(assetId)) {
+                /** @type{BuiltinAssetRecord} */
+                const assetRecord = typeBucket[assetId];
+                const asset =
+                    new Asset(assetRecord.type, assetRecord.id, assetRecord.format, new Buffer(assetRecord.data));
```
cwillisf: This creates new `Asset` and `Buffer` objects each time in case the client decides to modify their contents. If we require that the caller not modify these objects we could premake all the `Asset` objects and just return them as-is.

src/ScratchStorage.js
```diff
@@ -1,5 +1,129 @@
...
+        });
+    }
+
+    /**
+     * Fetch an asset and all its dependencies recursively.
+     * @param {AssetType} assetType - The type of asset to fetch.
+     * @param {string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
+     * @return {Promise.<Asset>} A promise for the requested Asset.
+     */
+    loadDeep (assetType, assetId) {
```
cwillisf: I'm starting to think that having this module handle the asset type and data format separately might be a mistake. What do you think?

The main reason I separated these concepts was for the sake of projects which bundle their assets, such as Scratch 1.4 projects and SB2 files, which I imagined would be automatically unpacked by `scratch-storage`'s `loadDeep` method into a structure something like this:

``` js
{
  assetType: AssetType.project,
  assetId: 1234,
  dataFormat: DataFormat.JSON,
  data: '{big JSON string of project data}'
  dependencies: [
    {
      assetType: AssetType.Image,
      assetId: 'aca2313b066d6253d626b61698819ef5',
      dataFormat: DataFormat.PNG,
      data: 'PNG....',
      dependencies: []
    },
    {
      // ...
    }
  ]
}
```

Having gone partway down that path, I'm starting to think that maybe `scratch-storage` should just handle opaque blobs of data and never parse / unpack them at all. In that case, the caller would deal with unpacking a Scratch 1.4 or SB2 project and its assets, or making additional requests for a Scratch 2.0 project JSON's assets.

Also, if we keep `loadDeep` I'm open to naming suggestions. `loadAssetThenParseForDependenciesAndFetchThemRecursively` seems a bit verbose.

src/DependencyParser.js
```diff
@@ -0,0 +1,49 @@
+const AssetType = require('./AssetType');
+const DataFormat = require('./DataFormat');
+
+class DependencyParser {
```
cwillisf: This class is just a stub for the sake of discussion. See [the comment on `loadDeep`](https://github.com/LLK/scratch-storage/pull/1/files#r87252090) below for why.

src/LocalHelper.js
```diff
@@ -0,0 +1,33 @@
+const Helper = require('./Helper');
+
+/**
+ * Implements storage on the local device, available even when the device has no network connection.
+ * TODO: use one or more of the following APIs for the web, and files under `app.getPath('userData')` for Node.
+ *       Maybe there's an existing Node module that would do that for us...?
```
cwillisf: LocalForage seems promising: https://github.com/localForage/localForage

src/WebHelper.js
```diff
@@ -0,0 +1,92 @@
+const Asset = require('./Asset');
+const Helper = require('./Helper');
+const xhr = require('xhr');
```
cwillisf: I suspect we want something fancier than `xhr` but this should get the idea across. I tried to use [`got`](https://www.npmjs.com/package/got) at first but had some trouble getting it to work with Webpack.

src/builtins/8e768a5a5a01891b05c01c9ca15eb6aa.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<svg width="128" height="128" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
+ <g>
+  <rect fill="#CCC" height="128" width="128"/>
+  <text fill="black" y="107" x="35.5" font-size="128">?</text>
+ </g>
+</svg>
```
cwillisf: This is equivalent to the `?` image below.

src/ScratchStorage.js
```diff
@@ -1,5 +1,129 @@
...
+     * Register a web-based source for assets. Sources will be checked in order of registration.
+     * @param {Array.<AssetType>} types - The types of asset provided by this source.
+     * @param {string} urlPattern - The URL pattern to use when loading assets from this source.
+     * The following sub-strings, if found in the URL pattern, will be replaced with properties of the requested asset:
+     * - '{ID}' will be replaced with the ID of the requested asset.
+     * - '{EXT}' will be replaced with the file extension of the requested asset (such as 'svg').
+     */
+    addWebSource (types, urlPattern) {
+        this.webHelper.addSource(types, urlPattern);
+    }
```
cwillisf, line 23: Examples:

``` js
// A project ID can be either '1234' or '1234:md5' where the latter specifies the exact revision of the project
addWebSource([AssetType.Project], asset => {
  const idParts = asset.assetId.split(':');
  idParts[1] = idParts[1] || '';
  return `https://cdn.projects.scratch.mit.edu/internalapi/project/${idParts[0]}/get/${idParts[1]}`;
});
addWebSource([AssetType.ImageBitmap, AssetType.ImageVector, AssetType.Sound], asset => {
  return `https://cdn.assets.scratch.mit.edu/internalapi/asset/${asset.assetId}.${asset.dataFormat}/get/`;
});
```

I imagine this growing such that the caller provides a way to retrieve not just a URL, but also any required headers, cookies, etc. for the sake of authentication.

src/ScratchStorage.js
tmickel: Hmmm...first, I'm trying to imagine what types of assets have dependencies except for projects. If there are not, I wonder if there would be any complexity reduction by making this a little less general-purpose.

My first feeling about asset types/data formats is to keep it similar to as-is: it would be nice to have scratch-storage unpack things. On the other hand, I don't know the SB1 format well enough to even know if this is feasible. For example, what _will_ we want to feed to the VM? I don't know the structure of the data well enough to know if it will be easy to separate the scripts/references to assets from the assets themselves.

Maybe it would be worth actually writing the loader (in Scratch VM, or scratch-storage) and then moving things back and forth depending on what we find? E.g., if it makes most sense to put the unpacking in the VM, then we can strip out the asset type/data format separation down the line.

src/Helper.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Base class for asset load/save helpers.
+ * @abstract
+ */
+class Helper {
```
tmickel, line 5: I wonder if we can find a better name than "helper."

src/DependencyParser.js
```diff
@@ -0,0 +1,49 @@
...
+     * @param {Asset} projectAsset - The project asset to parse.
+     * @return {Asset[]} - The list of dependencies. These may or may not contain their data.
+     */
+    static getProjectDependencies (projectAsset) {
+        switch (projectAsset.dataFormat) {
+        case DataFormat.SB2:
+            // TODO: unzip to retrieve JSON, then build dependency assets WITH data.
+            return projectAsset.dependencies;
+        case DataFormat.JSON:
+            // TODO: parse JSON for sprites, etc. WITHOUT data.
```
tmickel: yeah. on the one hand it would be nice to do it here. on the other hand this already happens in the VM (and pretty sure it needs to happen there to build objects). I guess it is related on your opinion on how dependent the VM and scratch-storage should be on each other. For example, can you imagine scratch-storage being useful in scenarios without a VM? In that case we may want unpacking to live here (and duplicated in the VM).

src/ScratchStorage.js
cwillisf: The other current example I can think of for an asset with dependencies is a sprite, which could have sounds and costumes as dependencies. We might want to automatically load a sprite's dependencies when loading it from a media library or from the backpack, for example.

A speculative example: we might want to be able to load a library of sprites or sounds or both, for example if we want to load a particular "media pack" into the editor. In this case the sprites and sounds could be considered dependencies. I don't think we'd want to automatically load all those dependencies, though.

I wonder if we should make another module for things like parsing a Scratch project (any version) and possibly other Scratch objects like sprites or media libraries. This module could be used by the VM and/or `scratch-storage`, and could also be used by external tools like the project recovery tool.

.eslintrc.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-    extends: ['scratch', 'scratch/node']
+    extends: ['scratch', 'scratch/es6', 'scratch/node']
```
rschamp: Take a look at scratch-gui for how to set up eslint so the node rules apply to the base project (for webpack, tests, etc.), but webpack/browser rules apply to what's in `/src`.

Base eslintrc: https://github.com/LLK/scratch-gui/blob/master/.eslintrc.js
src eslintrc: https://github.com/LLK/scratch-gui/blob/master/src/.eslintrc.js

src/BuiltinHelper.js
rschamp: If you do set it up as in `scratch-gui`, I think you won't have to disable global-require.

src/DependencyParser.js
cwillisf: Having thought about this for - *ahem* - a while now, I think it would be better for `scratch-storage` to not deal with dependencies or asset parsing at all. If you request a project with ID 1234 and that project happens to be in SB format, then you get an SB file and you're responsible for parsing and unpacking it.

I still think the concept of an asset with dependencies is a good one to use during project save, though. That will allow `scratch-storage` to ensure that every asset is saved before saving the project JSON, for example, but `scratch-storage` won't be responsible for ensuring consistency. For example, `scratch-storage` won't parse the project JSON to ensure that it refers to exactly the set of assets which are attached as dependencies.

.eslintrc.js
cwillisf: Since this is set up to use Webpack even for the browser-targeted build, should the stuff under `src/` specify Node-like linting? In particular, I use `Buffer` in the hopes that Node will take it as-is, and Webpack will convert it into `Uint8Array` for the web. Should I just use `Uint8Array` directly, or is there a better way to set this up?

rschamp: I'm not 100% but I don't think webpack will do that conversion for you unless you know about a plugin to help with it (maybe `babel-loader`?). Out of the box, webpack really only helps with `require`.  So I think using `Uint8Array` and assuming a browser environment in both cases (without allowing browser-exclusive globals) would be the simplest way to approach it.

cwillisf: I wasn't 100% sure about this stuff either, so I decided to investigate. It turns out that Webpack -- just plain Webpack, no `babel-loader` or anything needed -- includes this: https://github.com/webpack/node-libs-browser

The [`target`](https://webpack.github.io/docs/configuration.html#target) setting determines (among other things) whether or not `node-libs-browser` is included in your pack. For example, the `web` target includes it but the `node` target doesn't.

.eslintrc.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
+    root: true,
```
rschamp, line 2: Curious about this one — it's already the root.

src/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    env: {
+        node: false,
+        browser: false
+    },
+    extends: ['scratch/es6']
```
rschamp: I would expect root to be here, since you want to override the node linting and start doing web linting instead. Or maybe not since you're doing node + web linting.

.eslintrc.js
cwillisf: Doesn't this prevent `eslint` from going up further in the directory hierarchy, for example finding the `.eslintrc.js` of an app which holds `scratch-storage` in its `node_modules`?

rschamp: It would theoretically — what is the situation where you would be linting your dependencies though?

src/.eslintrc.js
cwillisf: It might make sense to set `root` here, yeah. It's turning out that this code is "minimally Node" ES6. In particular, it's very difficult to avoid `Buffer` completely, but I'm doing what I can to limit the parts of the code which have to know about `Buffer`. I definitely want to avoid anything that assumes we're on the web; I had to drop `arraybuffer-loader` because it uses `window.atob` which doesn't exist in Node.

.eslintrc.js
cwillisf: When I run `npm run lint` in `node_modules/scratch-storage/` which is `npm link`-ed to my dev repository :)

rschamp: Oh... I would say don't do that and lint from the dev repository! But this doesn't hurt anything so I say leave it if you're already working that way.

src/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    root: true,
+    extends: ['scratch', 'scratch/es6', 'scratch/node'],
+    env: {
+        node: false,
+        browser: false
+    }
```
cwillisf, line 7: This is a bit weird, but I think it's maybe the best option for the way I'm trying to treat this code?
Copied from the commit description:
> The code in `src/` is intended for use in both Node and on the web, but will be processed by Webpack which emulates a subset of Node's features when targeting the web. This change separates the linting rules for `src/` from the root of the project (which is intended for Node directly) and enforces Node-like linting rules without actually including Node globals. If we end up needing any of the Node globals provided by Webpack, we can add them explicitly here.

cwillisf: The `browser: false` bit isn't necessary but I wanted to be explicit that neither set of globals should be allowed here.

rschamp: That all makes sense to me!

PR #2, opened 2017-03-09
Get ready for use by scratch-vm & scratch-render

cwillisf:
These are changes that I made in the process of getting this module to work with scratch-vm and (indirectly) scratch-render.

Some of the more important changes:
- Upgrade Webpack from 1.x to 2.x
- Use `Buffer` instead of `Uint8Array` for asset data. These two types are very similar but it turns out that `Buffer` works better for consumption in scratch-vm & scratch-render.
- Add helper methods on `Asset` to convert to text or a data URI. To support this, each `AssetType` now knows its default MIME type.
- Don't reject a Promise unless there's an actual runtime error. It turns out that the debugger in some browsers treats a rejected Promise like a thrown exception, so using Promise rejection for control flow can be disruptive.
- Web build: export the storage module in the `Scratch` "namespace"

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "file-loader": "0.9.0",
     "got": "5.7.1",
     "json-loader": "0.5.4",
-    "localforage": "1.4.3",
+    "localforage": "1.5.0",
     "tap": "8.0.1",
     "text-encoding": "0.6.4",
     "webpack": "2.2.1"


src/.eslintrc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
     root: true,
     extends: ['scratch', 'scratch/es6', 'scratch/node'],
     env: {
-        node: false,
+        node: true,
```
cwillisf, line 5: This allows the use of `Buffer`

PR #6, opened 2017-03-21
Offer unpacked code from package.json

cwillisf:
### Proposed Changes

This change declares `./src/index.js` as the "main" script in `package.json`, and stops building `dist/node/*` with webpack. This change also restores the ES6 version of the test scripts, recently altered for Node.js 4.x compatibility.

### Reason for Changes

This sets us up to build `scratch-gui` in a single step, allowing webpack to better optimize our files and reducing the duplication of code within our output bundles.

### Test Coverage

Test coverage has not changed -- just the syntax of the test scripts.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,20 +1,25 @@
-const path = require('path');
 const webpack = require('webpack');
 
-const base = {
+module.exports = {
+    devtool: 'cheap-module-source-map',
+    entry: {
+        'scratch-storage': './src/index-web.js',
+        'scratch-storage.min': './src/index-web.js'
+    },
     module: {
         rules: [
             {
-                include: [
-                    path.resolve(__dirname, 'src')
-                ],
-                test: /\.js$/,
+                test: /[\\/]+scratch-[^\\/]+[\\/]+src[\\/]+.+\.js$/,
                 loader: 'babel-loader',
                 options: {
                     presets: ['es2015']
                 }
             }
         ]
+    },
+    output: {
+        path: __dirname,
+        filename: 'dist/web/[name].js'
     },
     plugins: [
         new webpack.optimize.UglifyJsPlugin({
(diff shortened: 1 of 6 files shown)


cwillisf:
Tests are broken due to the `require("binary-loader!...")` change. Will fix tomorrow...

webpack.config.js
```diff
@@ -1,21 +1,26 @@
-const path = require('path');
 const webpack = require('webpack');
 
-const base = {
+module.exports = {
+    devtool: 'cheap-module-source-map',
```
cwillisf, line 4: This option has treated me well over the past few days, though I'm happy to discuss and experiment with other options.
See also: https://webpack.js.org/configuration/devtool/#devtool

webpack.config.js
```diff
@@ -1,21 +1,26 @@
...
+        'scratch-storage.min': './src/index-web.js'
+    },
     module: {
         rules: [
             {
-                include: [
-                    path.resolve(__dirname, 'src')
-                ],
-                test: /\.js$/,
+                test: /[\\/]+scratch-[^\\/]+[\\/]+src[\\/]+.+\.js$/,
```
rschamp, line 12: Same comment as on the GUI — I think it would be more understandable if this was more explicit/verbose.

cwillisf:
Tests are working now, through the magic of jsdom.
Team: please give 2859875 some extra scrutiny to make sure you can stomach doing things this way :)

cwillisf:
The test diffs are more readable if you ignore white-space changes by adding `?w=1` to the end of the URL, by the way:
https://github.com/LLK/scratch-storage/pull/6/files?w=1

rschamp:
Whoops my comment didn't go through: set to needs work for the babel loader regex.

cwillisf:
Closing this in favor of #8 for now. There are a few other changes, like proper version numbers, which should happen before I try this kind of change again.

PR #9, opened 2017-04-21
Add `Asset` and `AssetType` to storage instance

cwillisf:
### Proposed Changes

Add `Asset` and `AssetType` to the `ScratchStorage` prototype, so that they're available on `ScratchStorage` instances. Also, deprecate accessing `Asset` or `AssetType` through `ScratchStorage` itself. These can be removed in a later change, once all consumers update to the new method of access.

The tests have also been updated to satisfy `eslint`'s `no-loop-func` rule.

### Reason for Changes

This allows consumers of this module to access `Asset` and `AssetType` without having to `require` the storage module itself. Specifically, this makes it possible to avoid building the `scratch-storage` module into both `scratch-vm` and `scratch-gui` separately.

### Test Coverage

The existing tests have been altered to access `Asset` and `AssetType` through a `ScratchStorage` instance.

### DIFF
--- a/src/ScratchStorage.js
+++ b/src/ScratchStorage.js
@@ -16,6 +16,40 @@ class ScratchStorage {
     }
 
     /**
+     * @return {Asset} - the `Asset` class constructor.
+     * @constructor
+     */
+    get Asset () {
+        return Asset;
+    }
+
+    /**
+     * @return {AssetType} - the list of supported asset types.
+     * @constructor
+     */
+    get AssetType () {
+        return AssetType;
+    }
+
+    /**
+     * @deprecated Please use the `Asset` member of a storage instance instead.
+     * @return {Asset} - the `Asset` class constructor.
+     * @constructor
+     */
+    static get Asset () {
+        return Asset;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


paulkaplan:
@cwillisf Looks like there are a few linting errors?

cwillisf:
Regarding linting:
Interesting... is `no-loop-func` a new rule or something? I can fix the issue but I'm surprised that it didn't come up before.

Regarding using accessors (@rschamp):
I think I'd like to keep the "static" access deprecated (and remove it in a little while) in order to discourage problematic instances of `require('scratch-storage')`, but moving things into the `ScratchStorage` class makes sense.

rschamp:
No it's been there since the first commit of eslint-config-scratch. I wonder if this is our decision to minor-pin eslint coming to bite us? Maybe there was a bug that got fixed in a new version, causing these to get caught when they weren't before?

SGTM re: static deprecation.

cwillisf:
@paulkaplan yep, will do! I actually have a branch for that going already, but I decided not to PR it just yet since I want to experiment a bit with the way we're constructing the storage object in the GUI.

PR #10, opened 2017-05-01
Prepare to use semantic-release

cwillisf:
### Proposed Changes

These changes will set us up to publish new versions according to the [Semantic Versioning](http://semver.org/) specification.

Specific changes:
- Install & configure `semantic-release` and `commitizen`.
- Configure Travis CI to use `semantic-release`.

### Reason for Changes

The `semantic-release` package will help us update our version number correctly; `commitizen` will help make commit messages that `semantic-release` will be able to parse.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,9 @@ install:
   - npm update
   - npm prune
 script:
-  - npm run test
+  - npm run lint
   - npm run build
+  - npm run tap
 after_script:
 - |
   # RELEASE_BRANCHES and NPM_TOKEN defined in Travis settings panel
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,12 @@ cache:
 node_js:
   - "4"
   - "node"
-before_install:
+install:
   - npm update
   - npm prune
+script:
+  - npm run test
+  - npm run build
 after_script:
 - |
   # RELEASE_BRANCHES and NPM_TOKEN defined in Travis settings panel
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
-language: node_js
-node_js:
-- "4"
-- "node"
 sudo: false
+language: node_js
 cache:
   directories:
-  - node_modules
-install:
-- npm install
-- npm update
+    - node_modules
+node_js:
+  - "4"
+  - "node"
+before_install:
+  - npm update
+  - npm prune
 after_script:
 - |
   # RELEASE_BRANCHES and NPM_TOKEN defined in Travis settings panel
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,14 +1,14 @@
...
-install:
-- npm install
-- npm update
+    - node_modules
+node_js:
+  - "4"
+  - "node"
+before_install:
+  - npm update
+  - npm prune
```
cwillisf: Moving `npm update` into `before_install` might help detect problems where a package update breaks the `npm run build` automatically triggered by `npm install`.

.travis.yml
```diff
@@ -23,9 +23,5 @@ after_script:
     # Don't release on PR builds
     $TRAVIS_PULL_REQUEST = "false"
   ]]; then
-    # Authenticate NPM
-    echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
-    # Set version to timestamp
-    npm --no-git-tag-version version $($(npm bin)/json -f package.json version)-prerelease.$(date +%s)
-    npm publish
+    npm run semantic-release
   fi
```
cwillisf, line 31: Do we still need `travis-after-all` or does `semantic-release` somehow handle that internally?

package.json
```diff
@@ -37,9 +39,18 @@
...
     "travis-after-all": "^1.4.4",
     "webpack": "2.2.1"
+  },
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
+  },
+  "release": {
+    "branch": "develop"
```
cwillisf, line 53: At some point in the future we may want to change this to "master" (or remove it, which is equivalent) but for now I think continuing to publish from "develop" makes sense.

.travis.yml
rschamp: Good idea using `npm prune`, since we're caching `node_modules/`.

I think we might see more confusing issues doing update in `before_install`, since then on install, if there are any issues with the build, we'll see them in the post install step of `npm install`. If we do the `npm update` after `npm install` we'll see those errors on the `npm build` step of the tests, which I think would be more intuitive.

I believe they fixed the behavior of running the `prepublish` script on `npm install`, so we can move that to `prepublishOnly` and then only build once per... build.

.travis.yml
rschamp: We still need travis-after-all AFAIK.

.travis.yml
cwillisf: Ah - I was specifically trying to ensure that the `npm update` happens before the build step, so if you're right about the `prepublish` change then that sounds ideal. I'll give it a shot.

.travis.yml
cwillisf: Cool; ~`eslint-config-scratch` might need a change then~.

Edit: nope. `estlint-config-scratch` only builds on one VM so that's why it doesn't need `travis-after-all`... uh... after all.

rschamp:
When you merge, I think you will need to correctly format the merge commit for a patch bump at least in order for semantic release to do its thing.

cwillisf:
Oh... I was going to let this be the last non-bumping merge and then future commits would use `cz`.

cwillisf:
For the record: I verified that this builds on Travis CI correctly both with and without a cache.

PR #11, opened 2017-05-04
Cache loaded assets with BuiltinHelper

rschamp:
### Resolves

Part of the resolution for:
- https://github.com/LLK/scratch-vm/issues/194
- https://github.com/LLK/scratch-vm/issues/550
- https://github.com/LLK/scratch-gui/issues/329

### Proposed Changes

- Add `cache` method to BuiltinHelper, to store retrieved assets
- Cache assets retrieved with other helpers with BuiltinHelper
- Flatten the BuiltinHelper cache structure, so only the md5 is needed to retrieve an asset. A collision between two types would be interesting but unlikely
- Add a synchronous `get` method to ScratchStorage that passes through to the BuiltinHelper for cached assets
- Fix WebHelper, so it sets the dataFormat on the Asset it returns
- Memoize base64 encoding of asset data by asset id

### Reason for Changes

This allows us to stop passing around asset data as a part of VM state, by storing the data in the storage instance.  Instead, consumers of the data only need an asset ID, and can retrieve it from the storage engine when required.

### DIFF
--- a/src/BuiltinHelper.js
+++ b/src/BuiltinHelper.js
@@ -70,6 +70,12 @@ class BuiltinHelper extends Helper {
         }
     }
 
+
+    /**
+     * Synchronously fetch a cached asset for a given asset id. Returns null if not found.
+     * @param {string} assetId - The id for the asset to fetch.
+     * @returns {?Asset} The asset for assetId, if it exists.
+     */
     get (assetId) {
         let asset = null;
         if (this.assets.hasOwnProperty(assetId)) {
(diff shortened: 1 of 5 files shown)


src/BuiltinHelper.js
```diff
@@ -83,23 +70,42 @@ class BuiltinHelper extends Helper {
...
+            const assetRecord = this.assets[assetId];
+            asset = new Asset(assetRecord.type, assetRecord.id, assetRecord.format, assetRecord.data);
+        }
+        return asset;
+    }
+
+    cache (assetType, dataFormat, data, id) {
+        if (!dataFormat) throw new Error('Data cached without specifying its format');
+        if (id) {
+            if (this.assets.hasOwnProperty(id)) return id;
```
cwillisf: This isn't safe for project files, where the same ID might change content over time. If it's helpful we could add a flag somewhere (on `AssetType`, for example) to indicate, per asset type, whether the asset ID is determined by its content or not.

cwillisf:
Oh, also: I noticed in the CI output that some of the "known asset" tests are failing to find their assets. The promises are getting rejected, meaning they encountered some error beyond just not finding the assets. Any idea why that is?

rschamp:
@cwillisf Fixed! It think it's a little weird for the id behavior to live with the asset type, but not sure where else to put it.

PR #12, opened 2017-05-24
feat(load): Allow caller to specify data format for an asset fetch

### LINKED ISSUE #7: scratch-storage incorrectly assumes all bitmaps are PNGs
### Expected Behavior

`scratch-storage` should be able to load JPG and GIF costumes, as present in the Scratch 2.0 costume library.

### Actual Behavior

`scratch-storage` assumes all bitmap costumes are PNGs, so it tries to load JPG and GIF costumes with the wrong exception. This results in a 404 and a failure to load the costume.

### Steps to Reproduce

1. Open your browser's network tab
2. Load project 152639677 in a version of `scratch-vm` or `scratch-gui` which uses `scratch-storage`
3. Look for a 404 on this address: https://cdn.assets.scratch.mit.edu/internalapi/asset/4f234884958b11023d44abc010ad50ec.png/get/

cwillisf:
### Resolves

Resolves #7

### Proposed Changes

When calling `load`, you may now explicitly state a data format / file extension. If you don't
specify one, the default for that asset type will be used.

### Reason for Changes

This allows loading assets whose file extensions don't match the default data format. For example, this allows loading JPEG-format backdrops.

### Test Coverage

I added several test cases to the "download known assets" integration test. These new cases ensure that we're covering `load` both with and without specifying the data format, as well as explicitly specifying a non-default format.

### DIFF
--- a/test/integration/download-known-assets.js
+++ b/test/integration/download-known-assets.js
@@ -16,6 +16,7 @@ test('constructor', t => {
  * @typedef {object} AssetTestInfo
  * @property {AssetType} type - The type of the asset.
  * @property {string} id - The asset's unique ID.
+ * @property {DataFormat} [ext] - Optional: the asset's data format / file extension.
  */
 const testAssets = [
     {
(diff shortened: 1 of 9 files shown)


test/integration/download-known-assets.js
```diff
@@ -82,12 +101,11 @@ test('load', t => {
     for (var i = 0; i < testAssets.length; ++i) {
         const assetInfo = testAssets[i];
 
-        const promise = storage.load(assetInfo.type, assetInfo.id);
+        var promise = storage.load(assetInfo.type, assetInfo.id, assetInfo.ext);
```
rschamp, line 104: Why not `let`?

cwillisf: This repo still uses Node 4, which is OK with `const` but not `let` in this context.

rschamp: TIL

PR #17, opened 2017-09-07
Use base64 package instead of toString('base64')

paulkaplan:
`toString` does not support arguments in the browser. Only in node (and apparently things got by got) allow an encoding to be passed to `toString`. 

This problem comes up when you try to use the `cache` function, adding your own data to storage instead of getting it from a URL. When it comes out of a URL (through `got`), a new `toString` function is overridden on the `Uint8Array` to make it act like a node `Buffer` object. When you try to cache your own `Uint8Array`, without that fancy toString, it fails to come back out correctly when you call `encodeDataURI`. 

_Aside: what the hell `got`?! Why are you overriding built-in methods of primitive JS objects. Very not cool. Damn near impossible to debug._ 

I added a new dependency that does this base64 conversion explicitly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "babel-loader": "6.2.10",
     "babel-polyfill": "6.22.0",
     "babel-preset-es2015": "6.22.0",
+    "base64-js": "1.2.1",
     "binary-loader": "0.0.1",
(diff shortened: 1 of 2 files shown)


paulkaplan:
@cwillisf yup the "won't be valid in any context" one!

Issue #16, opened 2017-09-06
Replace got


aeons:
`got` is a wrapper for node's `http` library. It currently does not work at all in an environment of an iOS webview, and the newest version seems to not work in browsers/webpack at all.

I suggest replacing it with something else. In our fork, I have replaced it with `whatwg-fetch` with great success, except the fact that tests fail, since it's a polyfill for browsers (`isomorphic-fetch` is outdated and doesn't allow for the same API across node/browser).

If you want, I can submit a PR with my changes.

thisandagain:
100% agreed! We would like to replace all occurrences of `got` with `nets` (which is an isomorphic wrapper on top of `request` and `xhr`). See:
https://github.com/LLK/scratch-vm/pull/585

aeons:
Sounds good, will PR tomorrow.

gengshenghong:
Got similar issue too! Also plan to replace with xhr way. As you started the worK, i will wait for your good news. Thank you.

aeons:
I tried adding nets, which was not really a pleasant experience. It's also a library that hasn't been updated for two years.

I tried adding node-fetch directly and using that, and that seems to work fine, both in the browser, in a web view and in tests. So I guess webpack takes care of making it run in the browser. Would you be interested in a PR replacing got with node-fetch?

thisandagain:
Hmm. @rschamp @cwillisf I'd like to normalize our decision here so it makes sense across repos. We are using `request` throughout a large number of repos without issue. I'm happy to move to something else but it should be consistent across all modules of 3.0 and (ideally) across all other projects (www, api, etc.). If we move this repository over to `whatwg-fetch` or `node-fetch` we will have three different HTTP request libraries in 3.0 which is not ideal.

rschamp:
@thisandagain I agree that whichever way we go for `scratch-storage`, all other 3.0 repos should follow.

@aeons If all of our other repos didn't already use `request`/`xhr`/`nets`, I think we would be interested in moving to some flavor of `fetch`, but as we are now, it'd be easiest to standardize on `nets`, and consider migrating to `fetch` at another time. Although it hasn't been updated in two years, `nets` is only a wrapper for `xhr` and `request`, which are individually upgradeable.

What was the unpleasantness you experienced with `nets`? Can I help work through the issues you had with it?

aeons:
[...]

PR #19, opened 2017-09-08
replace 'got' with 'nets' and fix the 'web+commonjs2' and 'node+commonjs2' targets in webpack.config.js

gengshenghong:
### Resolves

https://github.com/LLK/scratch-storage/issues/16

### Proposed Changes

1. replace 'got' with 'nets'
2. standardize the targets in webpack.config.js
web+umd
web+commonjs2
node+commonjs2
set 'web+commonjs2' as default 'main' filed in package.json

### Reason for Changes

See the issue for details.

### Test Coverage

Tested in 'scratch-gui' (web+commonjs2) and 'npm run tap' in 'scratch-storage' (node + commonjs2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "type": "git",
     "url": "https://github.com/LLK/scratch-storage.git"
   },
-  "main": "./dist/node/scratch-storage.js",
+  "main": "./dist/web_commonjs2/scratch-storage.js",
   "scripts": {
     "build": "./node_modules/.bin/webpack --progress --colors --bail",
     "coverage": "./node_modules/.bin/tap ./test/{unit,integration}/*.js --coverage --coverage-report=lcov",
@@ -38,6 +38,7 @@
     "json": "^9.0.4",
     "json-loader": "0.5.4",
     "localforage": "1.5.0",
+    "nets": "^3.2.0",
     "semantic-release": "^6.3.2",
     "tap": "8.0.1",
     "text-encoding": "0.6.4",
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -68,7 +68,7 @@ class WebHelper extends Helper {
                 if (urlFunction) {
                     const url = urlFunction(asset);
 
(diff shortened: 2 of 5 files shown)


cwillisf:
Thanks for working on this, @gengshenghong! Could you explain a bit more about the three targets you've set up? Here's what I understand:
- Node+CommonJS2 is used by tests
- Web+CommonJS2 is used by other Scratch 3.0 repositories importing this through Webpack
- Web+UMD is not currently used (?)

Is that understanding accurate, or is there something that uses (or will use) the web+UMD target?

At any rate, I really appreciate your work on this. We've been struggling for some time with the issues such as those you mention [here](https://github.com/LLK/scratch-storage/issues/16#issuecomment-328011802) -- not just with `nets` but with some of our own code as well. That was the main motivation for #9, for example. It's great to hear from someone who is a bit more familiar with these issues!

gengshenghong:
@cwillisf You are exactly right regarding the 3 targets I set up in webpack.config.js. I've also met into issues in one of my projects several weeks ago, and finally figured out the `libraryTarget` configuration will affect a lot.

Regarding web + UMD, it is not currently used in `scratch-storage`. Since we've setup that in current configuration in all scratch repos, I just keep it there. Also, it is useful in some cases, for example, in `scratch-vm`, the playground is using UMD (the very classical way to import library using <script></script>), to make things consistent, I'll suggest to keep it in all repos, when writing some browser tests, it will be much easier to import libraries using `<script></script>`.

**Clarify and Update**: the word "UMD" may be misunderstood by me, according to the document, the default libraryTarget is "var", not "UMD".
https://webpack.js.org/configuration/output/#output-librarytarget
I am not sure what it should be called for this case (I mean, import library using <script></script> in web browser...), if no better names, we may just call it `web+var` to follow webpack naming. I've only heard about UMD, CMD, AMD, commonJS2 systems, which are modular systems (I guess <script></script> usage is very traditional and not a kind of modular system at all).

Thanks,
Shenghong

PR #20, opened 2017-09-13
Use nets

### LINKED ISSUE #16: Replace got

cwillisf:
### Resolves

Resolves #16 and closes #19

### Proposed Changes

This is a slight refinement to & simplification of @gengshenghong's work in #19. There are several things going on here:
- I've added a `"browser"` property in `package.json`: this lets `webpack` use a different entry point compared to what `node` would use (specified in `"main"` property as usual). See this page for more details: https://github.com/defunctzombie/package-browser-field-spec
- The web output now generates a UMD module. This should not affect the Scratch team, but it makes the module consumable in more situations (including AMD/RequireJS) and has only minor overhead.
- Clean up path references in `webpack.config.js`, and in particular remove `__dirname`. I recently discovered that using `__dirname` in `webpack.config.js` can cause problems when symlinks are involved.
- Replace `got` with `nets`; this is @gengshenghong's work.

If this works well, I think we should apply the UMD and "browser" field changes to all our Scratch 3.0 repositories. In the longer term, I think we should consider optimizing our build outputs by splitting dependencies into a separate file ("vendor" output).

Thanks, @gengshenghong, for getting the ball rolling on this one!

### Reason for Changes

We'd like to standardize on `nets` as our XHR/request wrapper for all Scratch 3.0 repositories, which in the case of `scratch-storage` means replacing `got` with `nets`. Unfortunately this replacement was not straightforward due to our Webpack configuration. See the discussion on #16 and #19 for more details.

### Test Coverage

The existing tests in this repository cover the s/`got`/`nets`/ change -- in particular, the `download-known-assets` test. Integration tests in other repositories cover the Webpack changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "type": "git",
     "url": "https://github.com/LLK/scratch-storage.git"
   },
-  "main": "./dist/web_commonjs2/scratch-storage.js",
+  "main": "./dist/node/scratch-storage.js",
+  "browser": "./dist/web/scratch-storage.js",
   "scripts": {
     "build": "./node_modules/.bin/webpack --progress --colors --bail",
     "coverage": "./node_modules/.bin/tap ./test/{unit,integration}/*.js --coverage --coverage-report=lcov",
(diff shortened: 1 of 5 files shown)


gengshenghong:
Using UMD seems great to make 'browser' usage consistent. And it is great package.json support predefined field "browser"!

`scratch-vm` should also be updated to be consistent.

PR #26, opened 2017-11-14
Add mp3 to data formats

ericrosenbaum:
I'm parking this here because it is not clear if it's necessary... I'm using storage to load mp3 files for the music extension, but it appears to work without this change!


cwillisf:
If you're specifying `'mp3'` as a string, rather than `DataFormat.MP3` as an expression, then that should work fine. I do think we'll want this change at some point, possibly when we work on saving projects to the asset & project servers.

PR #29, opened 2018-02-13
Use commitlint to ensure the creation of new versions

rschamp:
This is an experiment to see if it would help to enforce commit message formatting through a linting tool.

On `npm install`, a git commit hook is created which checks that the format of your commit message conforms with the conventional changelog. If your message is malformatted, you're presented with a nice message that lets you know the issue.

This also includes a tool to do the same checking on Travis, for cases where a commit is created outside of an environment with commit hooks.  Supposedly this will lint all the commits for a PR — I guess we'll find out!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
   - npm update
   - npm prune
 script:
+  - commitlint-travis
   - npm run lint
   - npm run build
   - npm run tap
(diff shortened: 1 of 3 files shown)


rschamp:
It looks like commitlint-travis did indeed lint each of the commits for the PR!  So the PR build behavior differs from the normal behavior – when I pushed up the branch it only linted the head commit message.

PR #32, opened 2018-04-25
Skip to next source if a source responds with 404

### LINKED ISSUE #31: WebLoader does not try next source if request returns a 404 response
### Expected Behavior

If an asset request returns a 404 response the WebLoader should skip that source and try the next.

### Actual Behavior

The 404 response body is returned.

### Steps to Reproduce

```
class Storage extends ScratchStorage {
    constructor () {
        super();
        this.addWebSource(
            [this.AssetType.Project],
            asset => `https://example.com/${asset.assetId}`
        );
    }
}

const storage = new Storage();
storage.load(storage.AssetType.Project, "x", storage.DataFormat.JSON)
    .then(project => console.log("I did load this:", project))
    .catch(e => console.error("Loading failed"));
```

---

I'll try to submit a PR for this.

mraerino:
### Resolves

Fixes #31

### DIFF
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -70,7 +70,7 @@ class WebHelper extends Helper {
 
                     nets({url: url}, (err, resp, body) => {
                         // body is a Buffer
-                        if (err) {
+                        if (err || Math.floor(resp.statusCode / 100) !== 2) {
                             tryNextSource();
                         } else {
                             asset.setData(body, dataFormat);


src/WebHelper.js
```diff
@@ -70,7 +70,7 @@ class WebHelper extends Helper {
 
                     nets({url: url}, (err, resp, body) => {
                         // body is a Buffer
-                        if (err) {
+                        if (err || String(resp.statusCode).substr(0, 1) !== '2') {
```
cwillisf: I like the idea but I would prefer not to convert the integer into a `String`. How about either:
- check if `resp.statusCode` is between 200 and 299, inclusive, or
- check if the integer part of `resp.statusCode / 100` is exactly 2

mraerino: @cwillisf I fixed this 🙂

PR #33, opened 2018-05-07
chore(build): update to webpack 4

mzgoddard:
### Proposed Changes

- Update to webpack 4
- Update loaders to versions for webpack 4

### Reason for Changes

Sibling of https://github.com/LLK/scratch-vm/pull/1104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache:
   directories:
     - node_modules
 node_js:
-  - "4"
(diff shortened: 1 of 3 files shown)


mzgoddard:
The build failure is because travis is testing against Node 4 (as well as latest). More recent webpack versions dropped support for Node 4. We'll need to update the bottom version travis builds on for webpack.

PR #41, opened 2018-08-23
Add Storage.store and ability to store to the web

rschamp:
### Resolves

Towards https://github.com/LLK/scratch-gui/issues/2968

### Proposed Changes

- Remove LocalHelper
- Add SB3 DataFormat
- Replace fetch-only "source" concept with 2-way "store" concept
- Enhance `.addWebStore` so we can make `PUT`/`POST` requests to web stores
- Add `.store` method

### Reason for Changes

We need to be able to modify projects on the project server using storage. This is a step towards that goal.


src/BuiltinHelper.js
```diff
@@ -93,19 +95,35 @@ class BuiltinHelper extends Helper {
...
+     * @param {AssetType} assetType - The type of the asset to cache.
+     * @param {DataFormat} dataFormat - The dataFormat of the data for the cached asset.
+     * @param {Buffer} data - The data for the cached asset.
+     * @param {string} id - The id for the cached asset.
+     * @returns {string} The calculated id of the cached asset, or the supplied id if the asset is mutable.
+     */
+    store (assetType, dataFormat, data, id) {
         if (!dataFormat) throw new Error('Data cached without specifying its format');
-        if (id) {
+        if (id !== null && typeof id !== 'undefined') {
```
kchadha: Do we want to allow empty string IDs?

src/ScratchStorage.js
```diff
@@ -171,6 +183,27 @@ class ScratchStorage {
             tryNextHelper();
         });
     }
+
+    /**
+     * Store an asset by type & ID.
+     * @param {AssetType} assetType - The type of asset to fetch. This also determines which asset store to use.
+     * @param {DataFormat} [dataFormat] - Optional: load this format instead of the AssetType's default.
+     * @param {DataFormat} [data] - Optional: store this data, rather than the data in the asset
```
kchadha: Is this type annotation incorrect? Also is `data` actually optional?

src/ScratchStorage.js
```diff
@@ -171,6 +183,27 @@ class ScratchStorage {
...
+     * @param {?string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
+     * @return {Promise.<id, title>} A promise for asset metadata
```
kchadha: Is this type annotation correct? I thought `.<x,y>` was used to denote instances of parameterized type (e.g. for maps it would be the type of the keys and the values respectively)...

src/ScratchStorage.js
```diff
@@ -171,6 +183,27 @@ class ScratchStorage {
...
+     */
+    store (assetType, dataFormat, data, assetId) {
+        dataFormat = dataFormat || assetType.runtimeFormat;
+        return new Promise(
+            (resolve, reject) =>
+                this.webHelper.store(assetType, dataFormat, data, assetId)
+                    .then(body => {
+                        this.builtinHelper.store(assetType, dataFormat, data, body.id);
+                        return resolve(body);
```
kchadha, line 202: Is the `return` unnecessary here?

src/WebHelper.js
```diff
@@ -45,52 +65,110 @@ class WebHelper extends Helper {
...
+     * @param {AssetType} assetType - The type of asset to create or update.
+     * @param {?DataFormat} dataFormat - DataFormat of the data for the stored asset.
+     * @param {Buffer} data - The data for the cached asset.
+     * @param {?string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
+     * @return {Promise.<object>} A promise for the response from the create or update request
+     */
+    store (assetType, dataFormat, data, assetId) {
+        const asset = new Asset(assetType, assetId, dataFormat);
+        // If we have an asset id, we should update, otherwise create to get an id
+        const create = assetId === null || typeof assetId === 'undefined';
```
kchadha: Same question here about empty string assetId...

src/WebHelper.js
```diff
@@ -45,52 +65,110 @@ class WebHelper extends Helper {
...
+
+        // Use the first source with the appropriate asset type and url function
```
kchadha: source -> store?

src/WebHelper.js
```diff
@@ -1,36 +1,56 @@
...
+
 const Asset = require('./Asset');
 const Helper = require('./Helper');
 
 /**
  * @typedef {function} UrlFunction - A function which computes a URL from asset information.
  * @param {Asset} - The asset for which the URL should be computed.
- * @returns {string} - The URL for the asset.
+ * @returns {(string|object)} - A string representing the URL for the asset request OR an object with configuration for
+ *                              the underlying `nets` call (necessary for configuring e.g. authentication)
```
cwillisf, line 12: I'm OK with this for now, but it seems like we might need a `nets`-independent way to handle this kind of thing in the future.

src/WebHelper.js
```diff
@@ -45,52 +65,110 @@ class WebHelper extends Helper {
...
+
+        return new Promise((resolve, reject) => {
+            if (!store) return reject('No appropriate stores');
+
+            let reqConfig = create ? store.create(asset) : store.update(asset);
+            if (typeof reqConfig === 'string') {
+                reqConfig = {
+                    url: reqConfig
+                };
+            }
```
kchadha, line 155: You should probably have an `else { return reject(...); }` here...

kchadha (commented):
In general, LGTM. I have a few nits and questions in the comments.

src/BuiltinHelper.js
cwillisf: +1, especially considering that the empty string is converted to an "Untitled #"-type title in Scratch 2.0

src/ScratchStorage.js
cwillisf: I found this example on the JSDoc web site: `@return {Promise<string>}`

I think the proper way to do this might be something like: `@return {Promise<{id: string, title: string}>}` (or is the `id` a `Number`?)

cwillisf (commented):
Similar to @kchadha, I think this looks pretty good with a few nits & questions.

src/ScratchStorage.js
rschamp: The stuff in the returned object could be anything the web store returns so I think `Promise.<object>` is what we want.

src/ScratchStorage.js
rschamp: By convention you should `return` when you call the callback so that you don't accidentally do stuff after it (like calling the callback a second time). Not strictly necessary in terms of code though.

src/WebHelper.js
rschamp: This is for backwards compatibility from when we only returned strings for urls. So now it converts those results into objects, setting the url attribute to that result.

src/WebHelper.js
rschamp: Agreed. Could be complicated though.

rschamp: E.g., here's how we have to do it atm (and the reason for this change):
```js
    getProjectCreateConfig () {
        return {
            url: `${this.projectHost}/`,
            headers: {
                'Content-Type': 'application/x-www-form-urlencoded'
            },
            withCredentials: true,
            json: true
        };
```

PR #43, opened 2018-08-28
Small fixes to #41

rschamp:
This is cleanup for #41, added during integration.

`nets` always returns an arraybuffer unless you give it the option `encoding: undefined`(??) So do that, so we get parsed JSON back.

Also return the asset id even if the response wouldn't return it as a part of a `put` request. I had this when we paired on it @cwillisf, but somehow it went missing.

### DIFF
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -155,7 +155,8 @@ class WebHelper extends Helper {
             }
             return nets(Object.assign({
                 body: data,
-                method: method
+                method: method,
+                encoding: undefined // eslint-disable-line no-undefined
             }, reqConfig), (err, resp, body) => {
                 if (err || Math.floor(resp.statusCode / 100) !== 2) {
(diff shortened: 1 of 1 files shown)


src/WebHelper.js
```diff
@@ -155,13 +155,14 @@ class WebHelper extends Helper {
             }
             return nets(Object.assign({
                 body: data,
-                method: method
+                method: method,
+                encoding: undefined // eslint-disable-line no-undefined
```
cwillisf, line 159: We always want this because while the data we send using `store` varies in format, the response is always (expected to be) JSON. Is the right way to think about this?

cwillisf (approved):
👍

src/WebHelper.js
rschamp: Right! Also if not, we can set `responseType` to `arraybuffer` when setting up a store that needs arraybuffer.

PR #44, opened 2018-09-26
fix: use js implemented encoding APIs if unavailable

mzgoddard:
### Resolves

Improves performance where binary buffers are decoded into utf8 or utf8 is encoded into binary. This improves general VM loading time and GUI uses of decoding buffers into strings like decoding to render SVG in the paint editor and decoding SVG to render in the sprite item selectors.

### Proposed Changes

Use browser implemented TextDecoder and TextEncoder when available. Otherwise use JS implemented TextDecoder and TextEncoder.

### Reason for Changes

Chrome >= 38, Firefox >= 20, Safari >= 10.3 and Opera >= 25 provide browser implemented TextDecoder and TextEncoder implementations. These implementations are 2 - 10 times faster than the currently used JavaScript implementation.

### Test Coverage

The improvement can be seen in Chrome with project 247578672 while the paint editor is open on the `cover` sprite.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -1,5 +1,17 @@
-const TextDecoder = require('text-encoding').TextDecoder;
-const TextEncoder = require('text-encoding').TextEncoder;
+// Use JS implemented TextDecoder and TextEncoder if it is not provided by the
+// browser.
+let _TextDecoder;
+let _TextEncoder;
+const encoding = require('text-encoding');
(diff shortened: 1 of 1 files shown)


mzgoddard:
Edge is currently developing their implementation: https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/

mzgoddard:
Depends on #45.

cwillisf (approved):
As long as #45 fixes the CI build, this looks good to me!

PR #46, opened 2018-10-04
fix: WebHelper returns object with id on project creation

benjiwheeler:
At the bottom of WebHelper:store(), the `body` parameter appears to be a JSON string, not a JSON object:

screen shot 2018-10-03 at 9 59 07 pm

Object.assign does currently convert the string to an object, but it does it character by character:

screen shot 2018-10-03 at 10 00 11 pm

The code here that was setting the id does succeed in doing that -- as long as the assetId is already set:

(If it's an integer value, setting it succeeds:)
screen shot 2018-10-03 at 10 00 51 pm

(...but if we're relying on the number being extracted from body, that doesn't work:)
screen shot 2018-10-03 at 10 01 04 pm

So it seems what we really need to do is to JSON.parse the body -- at least before we try to access its content-name property. 

Existing behavior:
PUT (updating project with id) response:
screen shot 2018-10-03 at 11 00 19 pm

POST (creating new project) response:
screen shot 2018-10-03 at 11 02 06 pm

Behavior with this PR:
PUT (updating project with id) response:
screen shot 2018-10-03 at 10 46 35 pm

POST (creating new project) response:
screen shot 2018-10-03 at 10 54 55 pm

Question: should we just do JSON.parse for all responses? Seems like it should be consistent.

### DIFF
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -161,9 +161,13 @@ class WebHelper extends Helper {
                 if (err || Math.floor(resp.statusCode / 100) !== 2) {
                     return reject(err || resp.statusCode);
                 }
-                return resolve(Object.assign({
-                    id: body['content-name'] || assetId
-                }, body));
+                const bodyObj = typeof body === 'string' ? JSON.parse(body) : body;
+                return resolve(bodyObj.id ?
+                    body :
+                    Object.assign(bodyObj, {
+                        id: bodyObj['content-name'] || assetId
+                    }
+                ));
             });
         });
     }


rschamp (changes requested):
@benjiwheeler Sorry about this, you hit it because we sort of just hacked in PUTs/POSTs, and it's in an intermediate hacky state. As we discussed offline, this issue stems from a misconfiguration in GUI combined with the hacky way that `Storage.store` interacts with the body returned from `WebHelper.store`.

When we refactor the VM to hold all the assets, we won't need to do this anymore:
https://github.com/LLK/scratch-storage/blob/0fddecef584892a3b877b525195ac0d9d65b8948/src/ScratchStorage.js#L201

With that removed, we won't need the hacky transformation of `content-name` into `id` here, and can push the handling of `body` up to GUI (or wherever else).

So for now we should just configure GUI so an object is returned by `nets` like storage is expecting here. Even though it should probably expect nothing.

rschamp:
Closing as we discussed.

PR #48, opened 2018-10-10
fix: attempt to parse all string responses as JSON

### LINKED ISSUE #47: WebHelper returns strange object after project json PUT; creating new project using POST does not return id
At the bottom of WebHelper:store(), the `body` parameter appears to be a JSON string, not a JSON object:

screen shot 2018-10-03 at 9 59 07 pm

Object.assign does currently convert the string to an object, but it does it character by character:

screen shot 2018-10-03 at 10 00 11 pm

The code here that was setting the id does succeed in doing that -- as long as the assetId is already set:

(If it's an integer value, setting it succeeds:)
screen shot 2018-10-03 at 10 00 51 pm

(...but if we're relying on the number being extracted from body, that doesn't work:)
screen shot 2018-10-03 at 10 01 04 pm

So it seems what we really need to do is to JSON.parse the body -- at least before we try to access its content-name property. 

Existing behavior:
PUT (updating project with id) response:
screen shot 2018-10-03 at 11 00 19 pm

POST (creating new project) response:
screen shot 2018-10-03 at 11 02 06 pm

Behavior should be:
PUT (updating project with id) response:
screen shot 2018-10-03 at 10 46 35 pm

POST (creating new project) response:
screen shot 2018-10-03 at 10 54 55 pm

benjiwheeler:
resolves: https://github.com/LLK/scratch-storage/issues/47

### DIFF
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -161,6 +161,17 @@ class WebHelper extends Helper {
                 if (err || Math.floor(resp.statusCode / 100) !== 2) {
                     return reject(err || resp.statusCode);
                 }
+                // xhr makes it difficult to both send FormData and automatically
+                // parse a JSON response. So try to parse everything as JSON.
+                if (typeof body === 'string') {
+                    try {
+                        body = JSON.parse(body);
+                    } catch (parseError) {
+                        // If it's not parseable, then we can't add the id even
+                        // if we want to, so stop here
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Good to merge?

rschamp (approved):
Hey this looks pretty good to me! I wonder why?

PR #50, opened 2018-10-22
feat: replace cache with createAsset. Deprecate cache

rschamp:
### Resolves

Towards https://github.com/LLK/scratch-vm/issues/1577

### Proposed Changes

Add a new convenience method `createAsset` which returns `Asset` objects to replace `storage.cache`. This PR deprecates `cache` and adds a warning, but behavior should be unchanged. 

To replace `cache`'s id-creation functionality, add a new `generateId` argument to the Asset constuctor, `Asset#setData` and `Asset#encodeTextData` (the methods which change the data of the asset, and therefore the md5 hash of the data). If received, the asset will recalculate its ID to be the md5 hash of its contents.

### Reason for Changes

Towards incremental saving, we want to start storing the `Asset` instances on the costume and sound objects in the VM rather than in a single store in storage. This will allow us to easily keep track of which assets have been updated since the last save.

This should also prevent something that should look like a memory leak, because currently unused assets stay referenced by `storage` and are never garbage collected. Replacing `Asset` instances on the costumes as they're updated will allow us to only ever hold on to assets that are in use by the project.

### DIFF
--- a/src/ScratchStorage.js
+++ b/src/ScratchStorage.js
@@ -68,7 +68,7 @@ class ScratchStorage {
     }
 
     /**
-     * Cache an asset for future lookups by ID.
+     * Deprecated API for caching built-in assets. Use createAsset.
      * @param {AssetType} assetType - The type of the asset to cache.
      * @param {DataFormat} dataFormat - The dataFormat of the data for the cached asset.
      * @param {Buffer} data - The data for the cached asset.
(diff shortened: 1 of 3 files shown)


cwillisf (approved):
LGTM... any thoughts on how project IDs fit into this?

src/ScratchStorage.js
```diff
@@ -154,7 +169,7 @@ class ScratchStorage {
                                 } else {
                                     // TODO? this.localHelper.cache(assetType, assetId, asset);
                                     if (helper !== this.builtinHelper && assetType.immutable) {
-                                        asset.assetId = this.builtinHelper.cache(
+                                        asset.assetId = this.builtinHelper._store(
```
rschamp, line 172: In a follow-up we should add a breaking change and remove this behavior altogether.

rschamp:
Project IDs will be supplied by the VM or GUI when a project json asset is created. I think when we remove all of the deprecated methods we can also get rid of the "immutable" flag, and just rely on the consumer of storage to determine when IDs should be generated, or when they should be supplied.

PR #54, opened 2018-10-24
feat(webhelper): parse urlOptions to determine remix, copy params

benjiwheeler:
For remixing and saving as a copy, we need to correctly query projects and scratchr2 endpoints so as to create projects with references back to parent projects, and with titles that reflect the nature of the new project. This change allows for an object of key-value pairs to be passed, which is parsed into url querystring parameters, keeping any existing querystring parameters intact.

### Resolves

https://github.com/LLK/scratch-gui/issues/3197

### DIFF
--- a/src/ScratchStorage.js
+++ b/src/ScratchStorage.js
@@ -190,13 +190,14 @@ class ScratchStorage {
      * @param {?DataFormat} [dataFormat] - Optional: load this format instead of the AssetType's default.
      * @param {Buffer} data - Data to store for the asset
      * @param {?string} [assetId] - The ID of the asset to fetch: a project ID, MD5, etc.
+     * @param {?object} urlParams - object of params to add to querystring
      * @return {Promise.<object>} A promise for asset metadata
      */
-    store (assetType, dataFormat, data, assetId) {
+    store (assetType, dataFormat, data, assetId, urlParams) {
         dataFormat = dataFormat || assetType.runtimeFormat;
         return new Promise(
             (resolve, reject) =>
-                this.webHelper.store(assetType, dataFormat, data, assetId)
+                this.webHelper.store(assetType, dataFormat, data, assetId, urlParams)
                     .then(body => {
                         this.builtinHelper.store(assetType, dataFormat, data, body.id);
                         return resolve(body);
--- a/src/WebHelper.js
+++ b/src/WebHelper.js
@@ -5,6 +5,25 @@ const log = require('./log');
 const Asset = require('./Asset');
 const Helper = require('./Helper');
 
+// takes a url and adds query string params, maintaining existing ones.
+// note that it url encodes incoming param values.
+const addUrlParams = (url, paramsToAdd) => {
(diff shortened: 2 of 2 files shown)


cwillisf (commented):
In #55 I'm adding a way for a `scratch-storage` consumer to add multiple `WebHelper` (or other Helper) instances; do you think that would help you here?

Also, I don't see a way to provide `urlParams` at the `ScratchStorage` layer; do you intend the caller to own a `WebHelper` instance separate from `ScratchStorage`, or maybe to access the `webHelper` property of `ScratchStorage` directly?

benjiwheeler:
(Sorry Chris, I should have been clearer in my description that this wasn't totally done, I had just wanted to get Ray's thoughts on if this direction made sense! But now it's actually in good shape -- I did indeed add urlParams to ScratchStorage.)

Re: #55, I think I need someone to explain the *Helper concept used in Storage to me. I mean I can pull the threads and figure out what's happening, but I don't really get *why* it's happening...

If I understand WebHelper, its job is to help scratchstorage instances know that they should be looking to particular web endpoints to GET/PUT/POST, and what params / xhr options to use to do that. Seems to me like this change -- to expand WebHelper so it can be passed an endpoint and an object of url params, and assemble them before making the request -- fits within WebHelper's existing wheelhouse.

I guess an alternative would be to make a helper that did this param creation, and make it separate from the base one? Or maybe even to make RemixHelper and SaveHelper etc.?

src/ScratchStorage.js
```diff
@@ -190,13 +190,14 @@ class ScratchStorage {
      * @param {?DataFormat} [dataFormat] - Optional: load this format instead of the AssetType's default.
      * @param {Buffer} data - Data to store for the asset
      * @param {?string} [assetId] - The ID of the asset to fetch: a project ID, MD5, etc.
+     * @param {?object} urlParams - object of params to add to querystring
      * @return {Promise.<object>} A promise for asset metadata
      */
-    store (assetType, dataFormat, data, assetId) {
+    store (assetType, dataFormat, data, assetId, urlParams) {
```
rschamp, line 196: Since this is an argument to `store`, not `webHelper` (which may not exist on some storage instances!), then I think it would be prudent to make it more generalized to helper config, rather than specifically to `webHelper`.  Maybe it should be called `opts` or `helperOpts` or something, and it could be passed to the `store` method of every helper. Each helper can decide whether those options are relevant to it or not. So for this use-case we could pass in `opts` as `{urlParams: <what we would have been passing in as urlParams>}`.

rschamp (changes requested):
I think this could definitely be useful, although I think it should be more generalized. Before we go this way though, I think we should decide how we want separate helpers to be used vs. webstores. Also if we can sidestep the issue completely by moving remix,copy,title params into the POST data for create requests, I think I'd rather do that than complexifying storage more.

benjiwheeler:
Closing because we're going to try putting params into the POST request body, rather than the querystring. See:
https://github.com/LLK/scratch-projects/pull/83

PR #55, opened 2018-10-24
feat: create `addHelper` to add helper w/ priority

cwillisf:
### Proposed Changes

This change creates a new `addHelper` method, used to add a Helper to the storage module. Every registered Helper now has a priority value, and Helpers are consulted in descending priority order during a load.

### Reason for Changes

This allows `scratch-desktop` to inject its own Electron-specific Helper into the storage system. In the longer term I think this will also allow us to simplify the storage module, and in particular sets us up to eliminate some redundant code in `ScratchStorage.js` and `WebHelper.js`.

### Followup

If this PR is merged I will file these as followup tasks:

1. The `store` method should support multiple Helpers, consulted in priority order like in the load case.
   - The `webHelper` property on `ScratchStorage` should be removed.
   - The `store` method on a Helper should distinguish between "I don't handle this kind of asset" (perhaps returning `null`) and "I tried to handle this and failed" (perhaps returning and rejecting a Promise). In the former case we should move on to the next helper; in the latter we should probably fail the overall `store` call.
   - This may also involve moving some of the `AssetType` filtering logic from `WebHelper` into `ScratchStorage`.
2. The `WebHelper` should only handle one "store" and should not contain failover logic.
   - IMO it still makes sense to provide load, save, and update handlers as long as they all operate on the same "store" on the web.
   - Any code which currently adds multiple stores (say, one for projects and another for images & sounds) should be converted to add multiple `WebHelper` instances instead.
   - The `load` method on `WebHelper` should be simplified to remove `tryNextSource`: instead it should rely on `tryNextHelper` in `ScratchStorage`.

### Test Coverage

_Please show how you have added tests to cover your changes_


test/unit/helper-priority.js
```diff
@@ -0,0 +1,50 @@
...
+        return this.shouldSucceed ?
+            Promise.resolve(new storage.Asset(assetType, assetId, dataFormat, Buffer.from(this.label))) :
+            Promise.reject(`This is an expected failure from ${this.label}`);
+    }
+}
+
+test('addHelper', t => {
+    storage.addHelper(new FakeHelper('third', true), -50);
+    storage.addHelper(new FakeHelper('first', false), 50);
+    storage.addHelper(new FakeHelper('second', false), 0);
```
rschamp: In this test, there's no tests on the result of `addHelper`, so it's just testing that `addHelper` doesn't throw an error... Secretly though it's just setting things up for the `priorities` test. That can fail later on if you try to parallelize running tests.

Maybe this one should add these and observe the length of `storage._helpers`? Then do the same setup in `priorities` and test the order?

rschamp (dismissed):
The tests are a little hard to read and have state shared between them to function. Not a super big deal but would be nice to fix (as long as that's what's happening!)

test/unit/helper-priority.js
```diff
@@ -0,0 +1,50 @@
...
+    t.end();
+});
+
+test('priorities', t => {
+    t.deepEqual(logEntries, []);
+    return storage.load('fakeAssetType', 'fakeAssetId', 'fakeDataFormat').then(() => {
```
benjiwheeler: Are these strings supposed to refer to existing objects?

rschamp: No they're just strings that will work with the `FakeHelper` fixture above. It constructs an `Asset` using them directly (`Asset` doesn't check to see if they're defined, e.g. if `'fakeAssetType'` is an actual `AssetType`).

rschamp (approved):

PR #62, opened 2018-12-11
fix: load asset content

ktbee:
### Resolves

When I tried to use the default SVG asset in scratch-storage's `builtins` folder, I found that the `data` property just returned a string that was the file's path and name. This prevented me from getting the file's contents. 

### Proposed Changes

Disable the Webpack `file-loader` loader for these `require` statements. Webpack will still use the specified `arraybuffer-loader` loader for these files, but not using `file-loader` will give `arraybuffer-loader` the contents of the files rather than a string with the path and name of the file.

### DIFF
--- a/src/BuiltinHelper.js
+++ b/src/BuiltinHelper.js
@@ -24,7 +24,7 @@ const DefaultAssets = [
         format: DataFormat.PNG,
         id: null,
         data: new Buffer(
-            require('arraybuffer-loader!./builtins/defaultBitmap.png') // eslint-disable-line global-require
+            require('!arraybuffer-loader!./builtins/defaultBitmap.png') // eslint-disable-line global-require
         )
     },
     {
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
I'm OK with these changes but I'm confused about the need since there's no mention of `file-loader` in `webpack.config.js`. In the interest of my own learning: is `file-loader` somehow included in the default Webpack config, and if so is that documented somewhere? (I can't seem to find it...)

ktbee:
@cwillisf that's a good question. I believe the rule for loading SVGs with `file-loader` in the scratch-gui is what's affecting what the builtin assets return. This is the relevant line in the gui's Webpack configuration: https://github.com/LLK/scratch-gui/blob/055b2fc541c3a40a1d432a791144b883bb3ec4be/webpack.config.js#L115-L121

cwillisf:
@ktbee aha! That makes sense. Thanks!

ktbee:
Also here is the relevant Webpack documentation, for anyone that's not familiar with the `!` convention: https://github.com/webpack/docs/wiki/loaders#loader-order

kchadha (approved):
@ktbee thanks for linking to the documentation. LGTM!

PR #77, opened 2019-01-24
Only evaluate text-encoding on systems that need it

mzgoddard:
### Resolves

Load and evaluate the javascript for Scratch faster.

### Proposed Changes

Evaluate `text-encoding` once we know it is needed. Making this actually actually happen will need changes in other scratch packages using `text-encoding`.

### Reason for Changes

This lets us skip evaluating ~500kb of encoding indices. Evaluating less code lets us get to kicking of loading data and assets in less time since the page is loaded.

### Benchmark Data

I'll be holding off on producing a table for this. In part as the change is needed in other packages but also the change will be hard to measure. I'll link to a congregate table including a larger set of changes.

I might update this spot with some less comprehensive numbers.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -2,8 +2,12 @@
 // browser.
 let _TextDecoder;
 let _TextEncoder;
-const encoding = require('text-encoding');
 if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
+    // Wait to require text-encoding until we _know_ its needed. This will save
+    // evaluating ~500kb of encoding indices that we do not need to evaluate if
+    // the browser provides TextDecoder and TextEncoder.
+    // eslint-disable-next-line global-require
+    const encoding = require('text-encoding');
     _TextDecoder = encoding.TextDecoder;
     _TextEncoder = encoding.TextEncoder;
 } else {


src/Asset.js
```diff
@@ -2,8 +2,12 @@
 // browser.
 let _TextDecoder;
 let _TextEncoder;
-const encoding = require('text-encoding');
 if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
+    /* eslint global-require:0 */
```
cwillisf:
```suggestion
```

src/Asset.js
```diff
@@ -2,8 +2,12 @@
...
+    // Wait to require text-encoding until we _know_ its needed. This will save
+    // evaluating ~500kb of encoding indices that we do not need to evaluate if
+    // the browser provides TextDecoder and TextEncoder.
+    const encoding = require('text-encoding');
```
cwillisf, line 10:
```suggestion
    // eslint-disable-next-line global-require
    const encoding = require('text-encoding');
```

cwillisf (approved):
Looks good! As a future enhancement, do you think it would make sense to split `text-encoding` out into a separate JS file so that people who don't need it can skip the download too?
One nitpick: I made a couple suggestions for disabling the `global-require` linting rule in a way that's more consistent with how we tend to do it in other Scratch 3.0 repositories.

rschamp (approved):
LGTM — I agree with @cwillisf's suggestion too!

PR #78, opened 2019-01-24
Fetch assets with a web worker

mzgoddard:
### Resolves

Load assets faster.

### Proposed Changes

When it makes sense and is possible load assets in a web worker. As well use fetch instead of nets xhr wrapper when possible. If we can't fetch in a worker or fetch in the main thread, fallback to the nets wrapper.

### Reason for Changes

Loading assets in a web worker shifts asset loading overhead, switching between native and js. It also reduces the chance of the js blocking work happening on one of those assets loading. These behaviour changes greatly reduce the time between operating on loading the raw data of assets.

### Benchmark Data

Can be compared against the baseline in https://github.com/LLK/scratch-vm/pull/1944.

project id | platform | run 1 | run 2 | run 3 | run 4 | run 5 | avg
--------- | ------ | ---- | ---- | ---- | ---- | ---- | ----
169401431 | chrome | 7124 | 6821 | 7042 | 6598 | 6794 | 6875.8
169401431 | firefox | 5946 | 4139 | 3739 | 4098 | 4456 | 4475.6
169401431 | safari | 10614 | 7203 | 7190 | 7281 | 6974 | 7852.4
169401431 | book | 18692 | 16970 | 17106 | 17232 | 16304 | 17260.8
173918262 | chrome | 1860 | 1472 | 1325 | 1283 | 1178 | 1423.6
173918262 | firefox | 1312 | 1188 | 1184 | 1255 | 1155 | 1218.8
173918262 | safari | 2207 | 1030 | 1155 | 1061 | 1079 | 1306.4
173918262 | book | 5642 | 3054 | 2489 | 2527 | 2921 | 3326.6
173918262 | pi | 26373 | 17820 | 16502 | 18263 | 16711 | 19133.8

### DIFF
--- a/src/BuiltinHelper.js
+++ b/src/BuiltinHelper.js
@@ -152,9 +152,14 @@ class BuiltinHelper extends Helper {
      * Fetch an asset but don't process dependencies.
      * @param {AssetType} assetType - The type of asset to fetch.
      * @param {string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
-     * @return {Promise.<Asset>} A promise for the contents of the asset.
+     * @return {?Promise.<Asset>} A promise for the contents of the asset.
      */
     load (assetType, assetId) {
+        if (!this.get(assetId)) {
+            // Return null immediately so Storage can quickly move to trying the
+            // next helper.
+            return null;
+        }
         return Promise.resolve(this.get(assetId));
     }
 }
--- a/src/FetchWorkerTool.js
+++ b/src/FetchWorkerTool.js
@@ -158,6 +158,22 @@ class PublicFetchWorkerTool {
     get (reqConfig, options) {
         return this.inner.get(reqConfig, options);
     }
+
+    /**
+     * Is sending supported?
+     * @returns {boolean} Is sending supported?
+     */
+    get sendSupported () {
+        return false;
+    }
+
+    /**
+     * Send data to a server with a worker that uses fetch.
+     * @throws {Error} A not implemented error.
+     */
+    send () {
+        throw new Error('Not implemented.');
+    }
 }
 
 module.exports = PublicFetchWorkerTool;
--- a/src/FetchWorkerTool.js
+++ b/src/FetchWorkerTool.js
@@ -1,0 +1,163 @@
+/**
+ * Get and send assets with a worker that uses fetch.
+ */
+class PrivateFetchWorkerTool {
+    constructor () {
+        /**
+         * What does the worker support of the APIs we need?
+         * @type {{fetch:boolean}}
+         */
+        this._workerSupport = {
+            fetch: typeof fetch !== 'undefined'
+        };
+
+        /**
+         * A possible error occurred standing up the worker.
+         * @type {!Error}
+         */
+        this._supportError = null;
+
+        /**
+         * The worker that runs fetch and returns data for us.
+         * @type {!Worker}
+         */
+        this.worker = null;
+
+        /**
+         * A map of ids to fetch job objects.
+         * @type {object}
+         */
+        this.jobs = {};
+
+        try {
+            if (this.getSupported) {
+                // eslint-disable-next-line global-require
+                const FetchWorker = require('worker-loader?{"inline":true,"fallback":true}!./FetchWorkerTool.worker');
+
+                this.worker = new FetchWorker();
+
+                this.worker.addEventListener('message', ({data}) => {
+                    if (data.support) {
+                        this._workerSupport = data.support;
+                        return;
+                    }
+                    for (const message of data) {
+                        if (this.jobs[message.id]) {
+                            if (message.error) {
+                                this.jobs[message.id].reject(message.error);
+                            } else {
+                                this.jobs[message.id].resolve(message.buffer);
+                            }
+                            delete this.jobs[message.id];
+                        }
+                    }
+                });
+            }
+        } catch (error) {
+            this._supportError = error;
+        }
+    }
+
+    /**
+     * Is get supported?
+     *
+     * false if the environment does not workers, fetch, or fetch from inside a
+     * worker. Finding out the worker supports fetch is asynchronous and will
+     * guess that it does if the window does until the worker can inform us.
+     * @returns {boolean} Is get supported?
+     */
+    get getSupported () {
+        return (
+            typeof Worker !== 'undefined' &&
+            this._workerSupport.fetch &&
+            !this._supportError
+        );
+    }
+
+    /**
+     * Request data from a server with a worker using fetch.
+     * @param {{url:string}} reqConfig - Request configuration for data to get.
+     * @param {{method:string}} options - Additional options to configure fetch.
+     * @returns {Promise.<Buffer>} Resolve to Buffer of data from server.
+     */
+    get ({url}, options = {method: 'GET'}) {
+        return new Promise((resolve, reject) => {
+            // TODO: Use a Scratch standard ID generator ...
+            const id = Math.random().toString(16)
+                .substring(2);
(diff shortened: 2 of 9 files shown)


src/BuiltinHelper.js
```diff
@@ -152,9 +152,14 @@ class BuiltinHelper extends Helper {
      * Fetch an asset but don't process dependencies.
      * @param {AssetType} assetType - The type of asset to fetch.
      * @param {string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
-     * @return {Promise.<Asset>} A promise for the contents of the asset.
+     * @return {!Promise.<Asset>} A promise for the contents of the asset.
```
cwillisf: The `!` means "never null" -- did you mean to add `?`?
(I think `?` is optional for reference types -- implied unless `!` is present -- but I'm a little fuzzy on that.)
See also: http://usejsdoc.org/tags-type.html

```suggestion
     * @return {?Promise.<Asset>} A promise for the contents of the asset.
```

src/FetchWorkerTool.js
```diff
@@ -0,0 +1,191 @@
...
+
+        /**
+         * A map of ids to fetch job objects.
+         * @type {object}
+         */
+        this.jobs = {};
+
+        try {
+            if (this.getSupported) {
+                /* eslint global-require:0 */
```
cwillisf:
```suggestion
                // eslint-disable-next-line global-require
```

cwillisf (approved):
👍
I marked a couple minor suggestions but overall it looks reasonable to me.

One question: is there a reason to depend on `fetch` instead of using `nets` in the worker? Using `nets` in the worker would require a little bit more Webpack config but might simplify some of the code: the "tool" abstraction might become unnecessary for example. I don't feel strongly about it, though... just curious.

mzgoddard:
- Commit @cwillisf's suggestions
- Squash fixups

mzgoddard:
- Add benchmark data

src/ScratchStorage.js
```diff
@@ -177,49 +177,56 @@ class ScratchStorage {
...
+            }
+
+            // Nothing went wrong but we couldn't find the asset.
+            return Promise.resolve(null);
+        };
+
+        return tryNextHelper()
+            .then(asset => {
+                if (asset === null) {
+                    return tryNextHelper();
```
kchadha: Why is this necessary when `tryNextHelper` already tries to handle null?

kchadha (commented):
Had a question about the code in the comments, otherwise looks good to me

src/ScratchStorage.js
mzgoddard: You're right its not needed. Its part of [this](https://github.com/LLK/scratch-storage/pull/78/files#diff-ba10d9fbbe07eb4ce18d86eda3f7ca2eL191) original handler mostly unmodified. I'll remove it. (It might also cause an infinite loop?)

mzgoddard: To clarify it'll loop when it can't find anything to load.

mzgoddard:
- Remove the bit of code @kchadha pointed out

kchadha (approved):
LGTM

src/FetchWorkerTool.js
```diff
@@ -0,0 +1,191 @@
...
+     * Request data from a server with a worker using fetch.
+     * @param {{url:string}} reqConfig - Request configuration for data to get.
+     * @param {{method:string}} options - Additional options to configure fetch.
+     * @returns {Promise.<Buffer>} Resolve to Buffer of data from server.
+     */
+    get ({url}, options = {method: 'GET'}) {
+        // TODO: This is rather knowledgeable of what data urls look like. Can
+        // it be abstracted or less knowledgable while letting FetchTool do the
+        // job with data?
+        if (/\/\w+\/get\/$|\.json/.test(url)) {
```
rschamp: What is the behavior of this if e.g., someone is using the scratch gui with different project/asset servers configured with a different URL structure? I think it's a high priority to avoid having this kind of deep knowledge if it's possible.

mzgoddard: I'm going to update WebHelper to choose a tool based on assetType (one for Projects and one for everything else).

mzgoddard:
@rschamp @kchadha I changed WebHelper to have two tools, assetTool and projectTool. Passing a filter enum to ProxyTool, projectTool will use tools that are ready (and don't need to initialize like say a webworker). Let me know what you think.

src/FetchWorkerTool.js
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Is get supported?
+     *
+     * false if the environment does not workers, fetch, or fetch from inside a
+     * worker. Finding out the worker supports fetch is asynchronous and will
+     * guess that it does if the window does until the worker can inform us.
+     * @returns {boolean} Is get supported?
+     */
+    get getSupported () {
```
rschamp: This is confusing because it looks like a function name. Maybe `isGetSupported`?

src/FetchWorkerTool.js
```diff
@@ -0,0 +1,163 @@
...
+            };
+        })
+            .then(body => new Uint8Array(body));
+    }
+
+    /**
+     * Is sending supported? always false for FetchWorkerTool.
+     * @returns {boolean} Is sending supported?
+     */
+    get sendSupported () {
```
rschamp: Maybe `isSendSupported`?

src/FetchWorkerTool.worker.js
```diff
@@ -0,0 +1,54 @@
...
+ * Register a step function.
+ *
+ * Step checks if there are completed jobs and if there are sends them to the
+ * parent. Then it checks the jobs count. If there are no further jobs, clear
+ * the step.
+ */
+const registerStep = function () {
+    intervalId = setInterval(() => {
+        if (complete.length) {
+            postMessage(complete.slice(), complete.map(response => response.buffer).filter(Boolean));
```
rschamp: Could you please add a comment explaining the deal with transferring the buffer to the parent here?

src/ProxyTool.js
```diff
@@ -0,0 +1,70 @@
...
+     * @returns {Promise.<Buffer>} Resolve to Buffer of data from server.
+     */
+    get (reqConfig, options) {
+        let toolIndex = 0;
+        const nextTool = err => {
+            const tool = this.tools[toolIndex++];
+            if (!tool.getSupported) {
+                return nextTool(err);
+            }
+            if (!tool) {
```
rschamp, line 72: Shouldn't we check this existence before the `getSupported` property?

src/NetsTool.js
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Send data to a server with nets.
+     * @param {{url:string}} reqConfig - Request configuration for data to send.
+     * @param {*} data - Data to send.
+     * @param {string} method - HTTP method to sending the data as.
+     * @returns {Promise.<Buffer|string|object>} Server returned metadata.
+     */
+    send (reqConfig, data, method) {
+        return new Promise((resolve, reject) => {
+            /* eslint global-require:0 */
```
rschamp: Can we use `eslint-disable-next-line` here instead?

rschamp (changes requested):
@kchadha and I looked this over and think it looks good overall. We just have a few nits for readability/commenting.  Thanks!

src/ProxyTool.js
mzgoddard: I don't know how that happened. 😕

mzgoddard:
- Made requested changes.

rschamp (approved):
Thank you @mzgoddard!

PR #79, opened 2019-02-04
Immediately use next tool

mzgoddard:
### Resolves

Load data as soon as possible.

### Proposed Changes

Depends on #78 

- return null in FetchWorkerTool for data files
- initialize FetchWorkerTool.worker late

### Reason for Changes

- return null in FetchWorkerTool for data files
  This lets ScratchStorage, WebHelper, and ProxyTool iterate through its options until it hits FetchTool to load data files. This will let the fetch execute as soon as `storage.load` is called instead of delaying the fetch call.
- initialize FetchWorkerTool.worker late
  This can let fetching data happen even earlier by delaying evaluating the worker-loader, creating the Blob (for systems that support it), and create the worker.

### Benchmark Data

This provides a small but important improvement.

Pending ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "js-md5": "0.7.3",
     "minilog": "3.1.0",
     "nets": "3.2.0",
-    "text-encoding": "0.7.0"
+    "text-encoding": "0.7.0",
+    "worker-loader": "^2.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.1.2",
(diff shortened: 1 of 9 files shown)


cwillisf (approved):
👍

kchadha (approved):
LGTM

mzgoddard:
@kchadha Does this mean the PR it depends on is `good to merge`? #78

kchadha:
@mzgoddard, all PRs marked `good to merge`, `blocked`, and `needs-qa`, should hold off on being merged into develop until after the big smoke test tomorrow. They should all be included in the testing branch though.

mzgoddard:
I think changes to #78 make this change unnecessary.

PR #80, opened 2019-02-06
fix: encode buffers with btoa through a binary string

mzgoddard:
### Resolves

Load gui and projects faster.

### Proposed Changes

- Use native btoa to encode buffers into base64 if available.

### Reason for Changes

- Native btoa is faster and uses less memory (allocations are more manually than in js) than base64-js.

### Benchmark Data

Pending ...

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -12,14 +12,63 @@ if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
     _TextEncoder = TextEncoder;
 }
 
-const base64js = require('base64-js');
 const md5 = require('js-md5');
 
 const memoizedToString = (function () {
+    /**
+     * The maximum length of a chunk before encoding it into base64.
+     *
+     * 32766 is a multiple of 3 so btoa does not need to use padding characters
+     * except for the final chunk where that is fine. 32766 is also close to
+     * 32768 so it is close to a size an memory allocator would prefer.
+     * @const {number}
+     */
+    const BTOA_CHUNK_MAX_LENGTH = 32766;
+
+    /**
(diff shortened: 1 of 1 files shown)


cwillisf (commented):
Does this do the right thing for non-ASCII characters? In particular I'm curious what happens if the data buffer contains something that looks like a multi-byte UTF-8 character.

mzgoddard:
```
const data = new Uint8Array([0, 32, 64, 96, 128, 160, 192, 224, 255]);
```

node:

```
Buffer.from(data).toString('base64')
> 'ACBAYICgwOD/'
```

chrome inspector:

```
btoa(Array.from(data, v => String.fromCharCode(v)).join(''))
> 'ACBAYICgwOD/'
```

Does that seem about right @cwillisf?

cwillisf:
It took me a little while to figure out why this isn't a problem, but I was forgetting that JavaScript uses UTF-16 instead of UTF-8 internally. This code is safe since you're only dealing with 8-bit values, not 16-bit values.

Here's the kind of problem I was worried about, though you need to use UTF-16 surrogate pairs to see it and thus can't use `btoa`:
```js
data2 = new Uint16Array([0xD834, 0xDF06])
Array.from(data2, v => String.fromCharCode(v)).join('')
> "𝌆"
```

cwillisf (approved):
👍

kchadha:
@cwillisf, I couldn't tell if this PR should wait to be merged until after the internal smoke test (and thus be included in  the stuff we're testing there) or if you think it's a safe enough change that it can just be merged in? If the latter, you can remove the extra labels I added (and merge or let @mzgoddard merge).

PR #82, opened 2019-02-20
Support new json api

mzgoddard:
### Resolves

Faster project loading by making the initial data request earlier.

### Proposed Changes

Depends on https://github.com/LLK/scratch-storage/pull/79

Return null on `/\/\d+\/$` requests in FetchWorkerTool so a main thread tool can hopefully request the content immediately.

### Reason for Changes

The original test expression was for older apis used in older versions of gui and still used in other playgrounds like vm. This updated expression supports the request currently made by gui and www.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "js-md5": "0.7.3",
     "minilog": "3.1.0",
     "nets": "3.2.0",
-    "text-encoding": "0.7.0"
+    "text-encoding": "0.7.0",
+    "worker-loader": "^2.0.0"
   },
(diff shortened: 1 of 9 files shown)


rschamp:
I left a comment on #78, https://github.com/LLK/scratch-storage/pull/78#pullrequestreview-209301087

I think this PR by itself is fine, but the deep knowledge of our URL schemes stood out to me.

mzgoddard:
Once we make changes in this the PRs depended on by this PRs, I don't think this PR will be needed.

PR #90, opened 2019-04-19
fix: overwrite options to fetch like they are for nets

### LINKED ISSUE #88: PR to load assets in a web worker not accepted because storage is issuing puts on the asset server
@picklesrus discovered in the bug hunt that Scratch is issuing puts to update assets on the asset server when it should be using Posts.  

We shouldn't ever try to update the md5 hash because they are unique and changing the image should have a new md5 hash.  

The server doesn't accept puts, anyway and was sending back 405 errors while saving assets.

This came from testing https://github.com/LLK/scratch-gui/pull/4726, which we reverted on develop.

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-storage/issues/88

### Proposed Changes

- Support overwrite default options storage uses for fetch requests like we do with nets
- Change the *Tool get and send APIs to just take a Request object

### Reason for Changes

scratch-gui uses the web store url functions to overwrite the method for assets to always be post. To carry that value to the request we need to pass it in the options to fetch somehow.

### DIFF
--- a/src/FetchTool.js
+++ b/src/FetchTool.js
@@ -18,8 +18,8 @@ class FetchTool {
      * @param {{method:string}} options - Additional options to configure fetch.
      * @returns {Promise.<Uint8Array>} Resolve to Buffer of data from server.
      */
-    get ({url}, options = {method: 'GET'}) {
-        return fetch(url, options)
+    get ({url, ...options}) {
+        return fetch(url, Object.assign({method: 'GET'}, options))
             .then(result => result.arrayBuffer())
             .then(body => new Uint8Array(body));
     }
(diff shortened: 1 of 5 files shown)


mzgoddard:
I included `withCredentials` support. (usage in gui: [scratch-gui/src/lib/storage.js](https://github.com/LLK/scratch-gui/blob/develop/src/lib/storage.js#L67)) I'm not sure if we need it. Especially because I can't figure how `nets` uses it. `nets` uses `request` which uses `http` which webpack polyfills. I can't find where `nets` or `request` configure `http` to use `withCredentials`. (`withCredentials` is a boolean property of XMLHttpRequest objects.)

kchadha (approved):
LGTM! Thank you for fixing this!

PR #110, opened 2019-08-22
fix: return status code from fetch send failures like the nets tool

### LINKED ISSUE #100: 403 errors are swallowed by fetch tool POST
The asset server is configured to return a JSON body along with a 403, but this line of code is ignoring the fact that you can get both a response AND not be successful https://github.com/LLK/scratch-storage/blob/develop/src/FetchTool.js#L45

We should be rejecting that promise chain on 403s, but because of the way the fetch tool returns just the text response and not the status, and does not reject the response itself, there is no way for the promise chain to fail. 

/cc @cwillisf @BryceLTaylor @kchadha this is the reason that you do not get the "project failed to save" when you get authentication gets bumped by signing in to another computer. You can clearly see the 403s when you try to save, but they are being swallowed, and since nothing actually uses the response body we aren't catching the error elsewhere.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-storage/issues/100

### Proposed Changes

Reject calls to `storage.store` that fail through the fetch tools "send" method because of server errors. Use the [`response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) property to see if the send was successful, and reject with [`response.status`](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) if it is not ok.  

### Reason for Changes

This will make it possible for consumers of storage to know when storing an asset fails, and the reason for the failure.

### Test Coverage

_Please show how you have added tests to cover your changes_

Added a unit test for the fetch tool covering the successful and not successful usages

/cc @rschamp

### DIFF
--- a/src/FetchTool.js
+++ b/src/FetchTool.js
@@ -42,7 +42,10 @@ class FetchTool {
         return fetch(url, Object.assign({
             credentials: withCredentials ? 'include' : 'omit'
         }, options))
-            .then(result => result.text());
+            .then(response => {
+                if (response.ok) return response.text();
+                return Promise.reject(response.status);
+            });
     }
 }
 
--- a/test/unit/fetch-tool.js
+++ b/test/unit/fetch-tool.js
@@ -1,0 +1,29 @@
+const test = require('tap').test;
+
+const FetchTool = require('../../src/FetchTool');
+
+test('send success returns response.text()', t => {
+    global.fetch = () => Promise.resolve({
+        ok: true,
+        text: () => Promise.resolve('successful response')
+    });
+
+    const tool = new FetchTool();
+    
+    return tool.send('url').then(result => {
+        t.equal(result, 'successful response');
+    });
+});
+
+test('send failure returns response.status', t => {
+    global.fetch = () => Promise.resolve({
+        ok: false,
+        status: 500
+    });
+
+    const tool = new FetchTool();
+
+    return tool.send('url').catch(reason => {
+        t.equal(reason, 500);
+    });
+});


kchadha (commented):
@paulkaplan, this looks good to me, but should we standardize how we're checking for success across the different tools? E.g. see the corresponding `send` method in the nets tool: https://github.com/LLK/scratch-storage/blob/develop/src/NetsTool.js#L62-L64

paulkaplan:
@kchadha yeah the fetch.ok documentation indicates that it does the same thing (i.e. response in the 200 range) https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

cwillisf (approved):
LGTM... sorry this fell off my radar!

cwillisf:
@paulkaplan does this need more testing before it gets merged? I was talking to @adroitwhiz and he suggested that we should merge this to make it easier to migrate from nets/request to fetch.

paulkaplan:
I don’t think it needs anything else, but there is a patch in www or GUI
that is covers over this issue which needs to be removed. I’ll need to get
back to my machine to see where that is and Whether it’ll need to be
coordinated

paulkaplan:
@cwillisf  took a look, here is a quick recap just so we are on the same page.

### Recap of the issue
scratch-storage currently **resolves** `storage.store` calls even when the network requests involved return non-200 responses. This is an expectation mismatch between the network library layer (resolve for any server response, only reject when servers cannot be reached) and our storage layer (expectation to reject when asset cannot be saved).

### Current workaround 
scratch-gui has a workaround for this, [implemented here](https://github.com/LLK/scratch-gui/blob/develop/src/lib/project-saver-hoc.jsx#L238-L242), which checks the **text of the response** manually (that is all it has access too, not status code) and relies on our specific server behavior to sort successful/failure. The really important thing for GUI is to avoid marking assets "clean" that fail to save. Originally that code just relied on the promise rejecting and never getting to the "then", and that is still the behavior, just with an additional check. I think that means it is safe to change scratch-storage to reject promises the way gui originally expected it to, and we can remove those lines in GUI at any future time. 

### Next steps
If that all sounds right, here is a checklist of what I think needs to be done to get this into production.
- [ ] Create a reliable test case where an asset fails to save. This can be done in a few ways, but probably the easiest is to copy/paste a sound over and over until it gets past 250 seconds long (do not select any part of the sound, just alternate clicking "copy" and clicking "paste" to paste the whole sound at the end of the sound, doubling its length)
- [ ] Document that test case in the bughunt doc for when this gets merged, and check it on staging as soon as it goes up (don't wait for bughunt) so it can be removed if it is obviously broken.
- [ ] Create a tech-debt issue on GUI to remove https://github.com/LLK/scratch-gui/blob/develop/src/lib/project-saver-hoc.jsx#L238-L242 since they are not needed anymore and make our code specific to our servers (bad). 

Sorry for the wall of text, happy to chat about this if anything there is surprising/unclear.

cwillisf:
I tried making a big sound the way you suggested, but unfortunately the sound editor protected me (or protected the servers?) from that before I could get to the point of trying to save. So here's an artisinal hand-crafted SB3 file which fails to save due to a large WAV file -- just rename to SB3 and load it into the editor!
[big-sound.sb3.zip](https://github.com/LLK/scratch-storage/files/4588810/big-sound.sb3.zip)

PR #113, opened 2019-11-11
Remove NetsTool and nets dependency

adroitwhiz:
### Proposed Changes

This PR removes `NetsTool` and the dependency on the `nets` package.

### Reason for Changes

Regardless of whether it's running in Node or the browser, `nets` pulls in both the `request` library (which is [in maintenance mode](https://github.com/request/request/issues/3142)) and the `xhr` library. Both have many dependencies that add around ~100kb to the bundle size:
image

These include, but are not limited to, 13 different packages depended upon by a polyfill for `String.prototype.trim`, a function which is supported in all browsers released in the last 7 years.

According to MDN, the Fetch API is supported on all browsers that Scratch targets, making the `nets` fallback unnecessary:
image

image

To retain Node support (e.g. for running tests), I have added `node-fetch`, an implementation of the Fetch API for Node, as a dev-only dependency.

I also have [a branch which removes the `nets` dependency from scratch-vm](https://github.com/adroitwhiz/scratch-vm/commit/3fed542c0191406ad0ab9a9a6b61eb61547ad1d1), but I don't want to submit that until you approve of this one to avoid submitting too many potentially unwanted PRs.

### Test Coverage

Testing will have to be manually done across browsers.


cwillisf (approved):
👍

fsih:
Testing: load and save projects in many browsers

PR #114, opened 2019-11-11
Replace text-encoding package with fastestsmallesttextencoderdecoder

adroitwhiz:
### Proposed Changes

This PR replaces the [deprecated and unmaintained `text-encoding` package](https://www.npmjs.com/package/text-encoding), which polyfills the `TextEncoder` and `TextDecoder` APIs, with [`fastestsmallesttextencoderdecoder`](https://www.npmjs.com/package/fastestsmallesttextencoderdecoder).

### Reason for Changes

The main draw of `text-encoding` is that it supports text encodings other than UTF-8. However, we never use this functionality.

This entirely unused support for legacy encodings adds **500kb** of encoding data to the package. Replacing the polyfill with `fastestsmallesttextencoderdecoder`, which only supports UTF-8, removes this 500kb.

The `text-encoding` package is also pulled in by a few other Scratch repositories (scratch-vm, scratch-gui, and scratch-sb1-converter), and they don't use the non-UTF-8 functionality either. I'd remove it in all those repositories too, but I'd rather not "spam" pull requests.

I have not benchmarked load times. That may be worth looking into.

### Test Coverage

I have tested this in Edge, which is the only browser without native support for the Text Encoding API, and it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,9 @@
   "dependencies": {
     "arraybuffer-loader": "^1.0.3",
     "base64-js": "1.3.0",
+    "fastestsmallesttextencoderdecoder": "^1.0.7",
     "js-md5": "0.7.3",
     "minilog": "3.1.0",
-    "text-encoding": "0.7.0",
     "worker-loader": "^2.0.0"
   },
   "devDependencies": {
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -3,11 +3,9 @@
 let _TextDecoder;
 let _TextEncoder;
 if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
-    // Wait to require text-encoding until we _know_ its needed. This will save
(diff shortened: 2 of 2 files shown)


cwillisf:
What problems do you think this could cause if it doesn't work? It seems like this would primarily impact loading and saving of text assets, like project JSON for example, so maybe loading and saving projects on the web site and/or SB3 files should be tested in particular. Does that sound right to you?
TL;DR: Could you add testing steps to this?

cwillisf (approved):
Code looks good! Could you add testing info before merging? Thanks!

fsih:
Usages of text encoder/decoder:
- to fix quirks in .sb2 svgs to display properly in scratch 3
- to load all SVGs
- to inline fonts in SVGs to display correctly in thumbnails

fsih:
Testing: load an SB2 with SVG assets, and make sure that quirks are fixed correctly (text fonts are updated, viewbox is not messed up)
Make sure that the font is showing up in thumbnails

PR #169, opened 2020-09-04
fix(builtinhelper): use Buffer.from instead of deprecated constructor

### LINKED ISSUE #168: `new Buffer(...)` is deprecated
### Expected Behavior

Deprecated APIs should not be used.

### Actual Behavior

The file `BuiltinHelper.js` calls `new Buffer(...)` as part of establishing its default assets. This call has been deprecated since Node 6.0.0 and in some contexts causes a message similar to this:

> [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

Although in this case the actual code is not a security concern, I still think this should get a relatively high priority. This is because the message can sometimes be seen by the user (for example, if scratch-desktop is started from a terminal) and warns about security, which could worry users.

cwillisf:
### Resolves

Resolves #168 

### Proposed Changes

Use `Buffer.from(...)` instead of `new Buffer(...)`.

### Reason for Changes

The `Buffer` constructor has been deprecated since Node 6; this change uses `Buffer.from` instead. Doing so prevents a deprecation message from being displayed in certain contexts, such as when scratch-desktop is started from a terminal.

### Test Coverage

Covered by existing unit tests in `test/unit/load-default-asset.js`. I also inspected the default assets' `data` properties locally.

### DIFF
--- a/src/BuiltinHelper.js
+++ b/src/BuiltinHelper.js
@@ -23,7 +23,7 @@ const DefaultAssets = [
         type: AssetType.ImageBitmap,
         format: DataFormat.PNG,
         id: null,
-        data: new Buffer(
+        data: Buffer.from(
             require('!arraybuffer-loader!./builtins/defaultBitmap.png') // eslint-disable-line global-require
         )
     },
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Looks good. When this is compiled for web, I assume some web-compatible version of Buffer is used? or is this run at compile time?

PR #192, opened 2020-11-10
chore: update eslint version, modify hasOwnProperty calls

### LINKED ISSUE #191: Don't call hasOwnProperty directly on our objects #2682
There was a recent security warning that warned against calling property functions on your own objects that you assume have been inherited from Object.prototype. E.g., don't do obj1.hasOwnProperty(propName).

Instead, you can call the original function, e.g. Object.prototype.hasOwnProperty.call(obj1, propName).

benjiwheeler:
Resolves https://github.com/LLK/scratch-storage/issues/191

* updates versions of eslint, babel-eslint, and eslint-config-scratch
* modifies hasOwnProperty calls to be safer, in line with new lint rules


src/Asset.js
```diff
@@ -9,7 +9,6 @@ if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
     _TextDecoder = encoding.TextDecoder;
     _TextEncoder = encoding.TextEncoder;
 } else {
-    /* global TextDecoder TextEncoder */
```
benjiwheeler, line 12: re: this change, see https://github.com/eslint/eslint/issues/11979 and discussion at https://scratchteam.slack.com/archives/CS1144QCE/p1604958351210800

benjiwheeler:
Need to find reviewer

package.json
```diff
@@ -38,12 +38,12 @@
     "@commitlint/cli": "8.2.0",
     "@commitlint/config-conventional": "8.2.0",
     "@commitlint/travis-cli": "8.2.0",
-    "babel-eslint": "10.0.3",
+    "babel-eslint": "10.1.0",
     "babel-loader": "^8.0.4",
     "cz-conventional-changelog": "^3.0.2",
-    "eslint": "5.7.0",
-    "eslint-config-scratch": "5.1.0",
-    "eslint-plugin-react": "7.12.4",
```
rschamp, line 46: Is the addition of the major version range `^` for eslint-config-scratch and eslint-plugin-react but not eslint or babel-eslint intentional?

rschamp (approved):
This all looks reasonable to me. Just one point of confusion.

package.json
benjiwheeler: As I recall, some other repo of ours had that pattern -- but now I can't find it! I checked around, and it seems like we're generally using the major version caret for eslint, so I added it here.

PR #370, opened 2022-05-04
fix: polyfill 'fetch' for Node target

cwillisf:
### Proposed Changes

Configure Webpack to polyfill the `fetch` global with (effectively) `require('node-fetch').default` but only in the `target: 'node'` build.

### Reason for Changes

This allows `FetchTool` to work in versions of Node which don't provide `fetch`. The current `scratch-vm` integration tests are one important example of this problem.

### Test Coverage

Covered by existing tests in `scratch-storage` and `scratch-vm`. The `scratch-vm` integration tests were failing when run with `scratch-storage@1.3.5` but they now pass with this change. I also tested using Node 18.1.0, which does provide `fetch` globally.

Note that the `download-known-asset` test previously defined `global.fetch` itself, which fixed the `fetch` issue for that particular test but didn't help other situations (such as the tests in `scratch-vm`). Now that the fix is at the `webpack` level, the workaround in the `download-known-asset` test is no longer necessary and has been removed.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const {ProvidePlugin} = require('webpack');
 const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 const base = {
@@ -64,6 +65,11 @@ module.exports = [
             'js-md5': true,
             'localforage': true,
             'text-encoding': true
-        }
+        },
+        plugins: [
+            new ProvidePlugin({
+                fetch: ['node-fetch', 'default']
+            })
+        ]
     })
 ];
--- a/test/integration/download-known-assets.js
+++ b/test/integration/download-known-assets.js
@@ -1,7 +1,5 @@
 const md5 = require('js-md5');
 const test = require('tap').test;
-// Polyfill the fetch API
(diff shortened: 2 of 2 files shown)


webpack.config.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path');
+const webpack = require('webpack');
```
kchadha: Should this just be an import of `ProvidePlugin` specifically?

cwillisf: I would be OK with switching it to:
```suggestion
const ProvidePlugin = require('webpack').ProvidePlugin;
```
or
```suggestion
const {ProvidePlugin} = require('webpack');
```

...but since this plugin is built into `webpack` (unlike `uglify-webpack-plugin`) it can't just be `require('provide-webpack-plugin')` or anything like that.

I wrote it the way it is because that's the way the example code is written in the Webpack docs, but since there's only one thing currently coming from `webpack` I don't feel strongly about which one we use. Maybe I lean toward `const {ProvidePlugin}` a bit, now that I've written out the options. Thoughts?

kchadha (approved):
LGTM!

PR #388, opened 2022-05-17
Fix fetch tool error handling

kchadha:
### Resolves

Builds on and closes #363
Closes #156 

### Proposed Changes

This PR builds on top of #363.

In addition to the changes from that PR, these changes:
- Add special support for a 404 response while searching for the requested asset
- When all our sources are checked and the asset was nowhere to be found (e.g. all the web sources 404'd and did not have any other errors), resolve null for the requested asset.
- If the asset was not found but there were also errors with the tools we were using to request the asset, track and report

### Reason for Changes

- Fix an issue with asset loading where if the asset is missing, storage was resolving an html 404 page as the requested asset data.
- Also fixes an issue where we were not properly handling the case when the WebHelper has exhausted all the stores to check
- Make sure the `WebHelper.load` function is consistent with the documented API of `ScratchStorage.load`
- Make sure WebHelper returns a null Asset when the asset could not be found instead of a skeleton Asset with a null `data` field.

### Test Coverage

- #363 adds tests for 200 and 500 responses.
- added another FetchTool unit test for the 404 response

### DIFF
--- a/src/FetchWorkerTool.worker.js
+++ b/src/FetchWorkerTool.worker.js
@@ -51,6 +51,7 @@ const onMessage = ({data: job}) => {
     fetch(job.url, job.options)
         .then(result => {
             if (result.ok) return result.arrayBuffer();
+            if (result.status === 404) return null;
             return Promise.reject(result.status);
         })
         .then(buffer => complete.push({id: job.id, buffer}))
--- a/src/FetchWorkerTool.worker.js
+++ b/src/FetchWorkerTool.worker.js
@@ -49,7 +49,10 @@ const onMessage = ({data: job}) => {
     jobsActive++;
 
     fetch(job.url, job.options)
-        .then(response => response.arrayBuffer())
+        .then(result => {
+            if (result.ok) return result.arrayBuffer();
+            return Promise.reject(result.status);
+        })
         .then(buffer => complete.push({id: job.id, buffer}))
         .catch(error => complete.push({id: job.id, error}))
         .then(() => jobsActive--);
--- a/src/ScratchStorage.js
+++ b/src/ScratchStorage.js
@@ -168,7 +168,7 @@ class ScratchStorage {
      * @param {string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
      * @param {DataFormat} [dataFormat] - Optional: load this format instead of the AssetType's default.
      * @return {Promise.<Asset>} A promise for the requested Asset.
-     *   If the promise is resolved with non-null, the value is the requested asset or a fallback.
+     *   If the promise is resolved with non-null, the value is the requested asset.
      *   If the promise is resolved with null, the desired asset could not be found with the current asset sources.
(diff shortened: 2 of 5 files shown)


src/FetchWorkerTool.worker.js
```diff
@@ -49,7 +49,10 @@ const onMessage = ({data: job}) => {
     jobsActive++;
 
     fetch(job.url, job.options)
-        .then(response => response.arrayBuffer())
+        .then(result => {
+            if (result.ok) return result.arrayBuffer();
+            return Promise.reject(result.status);
+        })
```
cwillisf, line 56: Seems like this one should handle `result.status === 404` as well?

src/ScratchStorage.js
```diff
@@ -168,7 +168,7 @@ class ScratchStorage {
      * @param {string} assetId - The ID of the asset to fetch: a project ID, MD5, etc.
      * @param {DataFormat} [dataFormat] - Optional: load this format instead of the AssetType's default.
      * @return {Promise.<Asset>} A promise for the requested Asset.
-     *   If the promise is resolved with non-null, the value is the requested asset or a fallback.
+     *   If the promise is resolved with non-null, the value is the requested asset.
```
cwillisf, line 171: 😎&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<sub><sub>(also 😓)</sub></sub>

test/unit/fetch-tool.js
```diff
@@ -27,3 +29,47 @@ test('send failure returns response.status', t => {
...
+    global.fetch = () => Promise.resolve({
+        ok: false,
+        status: 500
+    });
+
+    const tool = new FetchTool();
+
+    return tool.get({url: 'url'}).catch(reason => {
+        t.equal(reason, 500);
+    });
```
cwillisf: I would suggest using `t.resolves` and `t.rejects` to test these promises... assuming that we're using a new enough `tap`. Right now I think this one would pass if the promise resolves (doesn't reject), since it won't get to the `t.equal` at all so there's no check to fail.

cwillisf: Update: I checked the `tap` version and it looks like we're good.

cwillisf (commented):
This is awesome! Thank you both, @AndreyNikitin and @kchadha, for helping resolve these long-standing storage issues!

I found a few quibbles but I think the concept is solid. Looking forward to this!

src/FetchWorkerTool.worker.js
kchadha: oh huh! idk how I missed that one

kchadha: updated this code!

cwillisf: 👍

cwillisf (approved):
Looks good!

PR #389, opened 2022-05-18
fix(fetchworkertool): don't create Uint8Array for null response

kchadha:
### Proposed Changes

Resolves issue missed by #388 where if FetchWorkerTool.worker resolves with null, FetchWorkerTool turns that into an empty Uint8Array. Resolve null instead.


cwillisf (approved):
🚤

PR #390, opened 2022-05-25
fix(worker): report worker errors as strings

cwillisf:
### Proposed Changes

If the `FetchWorkerTool` encounters an error, record it in the completion queue as a string, not an `Error` object.

### Reason for Changes

Error objects can't be sent with `postMessage`, so the previous code would throw an exception when trying to send the error back to the main thread.

### Test Coverage

Tested locally

### DIFF
--- a/src/FetchWorkerTool.worker.js
+++ b/src/FetchWorkerTool.worker.js
@@ -55,7 +55,7 @@ const onMessage = ({data: job}) => {
             return Promise.reject(result.status);
         })
         .then(buffer => complete.push({id: job.id, buffer}))
-        .catch(error => complete.push({id: job.id, error}))
+        .catch(error => complete.push({id: job.id, error: error.message || `Failed request: ${job.url}`}))
         .then(() => jobsActive--);
 };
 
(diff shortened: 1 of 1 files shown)


src/FetchWorkerTool.worker.js
```diff
@@ -55,7 +55,7 @@ const onMessage = ({data: job}) => {
             return Promise.reject(result.status);
         })
         .then(buffer => complete.push({id: job.id, buffer}))
-        .catch(error => complete.push({id: job.id, error}))
+        .catch(error => complete.push({id: job.id, error: error.message || `Failed request: ${job.url}`}))
```
kchadha:
```suggestion
        .catch(error => complete.push({id: job.id, error: (error && error.message) || `Failed request: ${job.url}`}))
```

Just in case we somehow reject `undefined` or `null` ¯\_(ツ)_/¯

src/FetchWorkerTool.worker.js
```diff
@@ -55,7 +55,7 @@ const onMessage = ({data: job}) => {
...
         .then(() => jobsActive--);
```
kchadha, line 59: Should this be a `.finally`?
```suggestion
        .finally(() => jobsActive--);
```

kchadha (approved):
Added minor suggestions but I don't think they are totally necessary.

Thanks for digging into this with me!

src/FetchWorkerTool.worker.js
cwillisf: I experimented with this and it doesn't seem to help. We should investigate further but for now I'll leave this off.

PR #404, opened 2023-03-09
Augment fetch

cwillisf:
### Resolves

This is a big step toward ENA-245

### Proposed Changes

Add a customized version of `fetch`, called `scratchFetch`, which can optionally include metadata as headers, then use `scratchFetch` throughout `scratch-storage`.

Supporting changes:

- Replace `node-fetch`, which only works in Node, with `cross-fetch`, which works in Node as well as browsers with and without native `fetch` support.
- Remove `ProvidePlugin` configuration to inject global `fetch` into the Node build, in favor of explicitly using `cross-fetch` in the places that need it.
- Add `@babel/plugin-transform-runtime` and `@babel/runtime` so I can use `async` and `await`.
- Add tests for `scratchFetch`, mocking `fetch` to avoid actual network connections.
- Update `tap` because our old version didn't support `tap.mock`.
- Add `"tap": {"check-coverage": false}` to `package.json` because newer `tap` has it turned on by default and we don't (yet?) meet their standards for coverage.

### Test Coverage

New tests cover all this fun new stuff!

Existing tests were updated to mock `fetch` more accurately, which actually exposed at least one error in a test (passing a bare URL string to a function expecting an object with a `url` property).

### DIFF
--- a/src/FetchTool.js
+++ b/src/FetchTool.js
@@ -9,12 +9,12 @@ const {scratchFetch} = require('./scratchFetch');
  */
 class FetchTool {
     /**
-     * Is get supported? false if the environment does not support fetch.
+     * Is get supported?
+     * Always true for `FetchTool` because `scratchFetch` ponyfills `fetch` if necessary.
      * @returns {boolean} Is get supported?
-     * @deprecated Fetch is always supported now
      */
     get isGetSupported () {
-        return typeof scratchFetch !== 'undefined';
+        return true;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/FetchTool.js
```diff
@@ -7,42 +11,43 @@ class FetchTool {
     /**
      * Is get supported? false if the environment does not support fetch.
      * @returns {boolean} Is get supported?
+     * @deprecated Fetch is always supported now
      */
     get isGetSupported () {
-        return typeof fetch !== 'undefined';
+        return typeof scratchFetch !== 'undefined';
```
aoneill01: Just personal preference, I think I would `return true;` here and `isSendSupported`. But this way does make it clearer why it is deprecated.

aoneill01 (approved):
I like all of your jsdoc type annotations. Looks good to me!

src/FetchWorkerTool.js
```diff
@@ -52,6 +54,8 @@ class PrivateFetchWorkerTool {
                         }
                     }
                 });
+
+                this.worker = worker;
```
colbygk, line 58: I like that this isolates the worker from the owning object until the worker has been fully constructed.

colbygk (approved):
LGTM!

PR #411, opened 2023-03-30
Improve scratchFetch metadata and testing

cwillisf:
### Proposed Changes

The main changes here are:
- transmit headers from `FetchWorkerTool.js` to `FetchWorkerTool.worker.js` as an array instead of a `Headers` object.
- expose `scratchFetch` through a getter on `ScratchStorage` instances

Also:
- store headers in a `Headers` object instead of a `Map`
- changes to the way the `fetch` function & `cross-fetch` module are mocked for tests

### Reason for Changes

While testing the `scratchFetch` in isolation, everything seemed good. Once I started testing it in `scratch-gui` setting a project ID, which only happens in the context of `scratch-www` connected to a project server, I started seeing errors when loading projects. Specifically, I was seeing CORS errors complaining about a header named `map`. I tracked this down to the `postMessage` in `FetchWorkerTool`: it was trying to send a `Headers` object instance, but `postMessage` doesn't support that. The `postMessage` was doing the best it could to turn the `cross-fetch` implementation of `Headers` into a plain object, which means it was getting a member called `map` based on its internal structure. The worker would then do as it was told and send a `map` header. So that's where the `Array.from()` comes from: that field is allowed to be a `Headers` object or anything that can be passed to the `Headers` constructor, which includes things like `{a:'A', b:'B'}` or `[['a','A'],['b','B']]`. I picked the array one because I'm (overly?) paranoid about stray properties in an Object.

Also in testing, I was running into trouble where two parts of the code (like a test in `scratch-gui` vs. code inside `scratch-vm`). I resolved this by adding a getter for `scratchFetch` on `ScratchStorage` so that all code sharing a particular instance of `scratch-storage` can share a single `scratchFetch` and its `metadata` object.

Storing headers in `Headers` object is just a nice-to-have meant to make `setMetadata` and `getMetadata` reflect what will actually happen with the headers. This means that `setMetadata('x-runid', ...)` will override `setMetadata('x-RunId', ...)`, for example. Also, if you try to set a header with an invalid character in its name, you'll get an error right away instead of the error being delayed until the next `scratchFetch` attempt.

I also ran into trouble testing because I mocked `cross-fetch` differently in two tests and missed that. I refactored the way I'm mocking `cross-fetch` so it's harder to make that mistake in the future.

### Test Coverage

Existing tests have been adjusted to cover the new structure, and I added one for case sensitivity in metadata keys.

### DIFF
--- a/src/FetchWorkerTool.js
+++ b/src/FetchWorkerTool.js
@@ -1,4 +1,4 @@
-const {applyMetadata} = require('./scratchFetch');
+const {Headers, applyMetadata} = require('./scratchFetch');
 
 /**
  * Get and send assets with a worker that uses fetch.
@@ -92,6 +92,13 @@ class PrivateFetchWorkerTool {
             const augmentedOptions = applyMetadata(
                 Object.assign({method: 'GET'}, options)
             );
+            // the Fetch spec says options.headers could be:
+            // "A Headers object, an object literal, or an array of two-item arrays to set request's headers."
+            // structured clone (postMessage) doesn't support Headers objects
+            // so turn it into an array of two-item arrays to make it to the worker intact
+            if (augmentedOptions && augmentedOptions.headers instanceof Headers) {
+                augmentedOptions.headers = Array.from(augmentedOptions.headers.entries());
+            }
             this.worker.postMessage({
                 id,
                 url,
--- a/src/ScratchStorage.js
+++ b/src/ScratchStorage.js
@@ -6,6 +6,7 @@ const WebHelper = require('./WebHelper');
 const _Asset = require('./Asset');
 const _AssetType = require('./AssetType');
 const _DataFormat = require('./DataFormat');
+const _scratchFetch = require('./scratchFetch');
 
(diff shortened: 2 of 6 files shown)


src/FetchWorkerTool.js
```diff
@@ -92,6 +92,13 @@ class PrivateFetchWorkerTool {
             const augmentedOptions = applyMetadata(
                 Object.assign({method: 'GET'}, options)
             );
+            // the Fetch spec says options.headers could be:
+            // "A Headers object, an object literal, or an array of two-item arrays to set request's headers."
+            // structured clone (postMessage) doesn't support Headers objects
+            // so turn it into an array of two-item arrays to make it to the worker intact
+            if (augmentedOptions && augmentedOptions.headers instanceof Headers) {
+                augmentedOptions.headers = Array.from(augmentedOptions.headers.entries());
```
aoneill01, line 100: That's a pretty line of code!

aoneill01 (approved):
Looks good. I appreciate the comments.

PR #412, opened 2023-05-02
Switch to Jest and re-enable project tests

cwillisf:
### Resolves

Resolves ENA-269

### Proposed Changes

Major changes:

* Convert all tests to use Jest
* Re-enable project tests in `download-known-assets`

Supporting changes:

* Fix a small mistake in the `load-default-assets` tests (see d4385b2fd18bf1da3f540813eda47cad2b454077)
* Move the specification of `arraybuffer-loader` from the `require` line into `webpack.config.js`
* Mock all network requests during tests

### Reason for Changes

* The current plan is to standardize on Jest
* Mocking all network requests means:
  * The project tests in `download-known-assets` can be re-enabled. They were disabled in #402 because they broke when the project server started requiring tokens.
  * These tests are now independent of our project servers and the CI server's Internet connection :)
* Tap has been giving me a lot of trouble lately, including with some modules using `import` instead of `require` and with `!loader!path/file` notation, but I really couldn't get it to work with the more complicated mocking situation I need to mock network requests for `download-known-assets`. Jest doesn't handle `!arraybuffer-loader!path/file` notation either, but it _does_ offer a different way to handle that (see `just.config.js`) and its mocking features made `download-known-assets` relatively easy.

Now that `download-known-assets` doesn't actually download from real servers, it may not be obvious why we need the test at all. I still think it's worthwhile, though. It shows that asset IDs make it through the system correctly in one direction and that asset data makes it back out correctly.

### Test Coverage

Tested locally on both Windows and Ubuntu. Backslashes are tricky.


test/unit/metadata.test.js
```diff
@@ -0,0 +1,123 @@
...
+beforeEach(() => {
+    // reset the metadata container to ensure the tests don't interfere with each other
+    // but this also means we need to `require` inside the tests
+    /* eslint-disable global-require */
+    jest.resetModules();
+});
+
+test('get without metadata', async () => {
+    const FetchTool = require('../../src/FetchTool.js');
+    const tool = new FetchTool();
```
delasare, line 12: Can these be moved to beforeEach or possibly a `const` at the top of the file?

delasare (approved):
One question - but this looks good to me!

test/unit/metadata.test.js
cwillisf: Each test starts with no modules loaded due to the `jest.resetModules()` in `beforeEach`, so they need to `require` everything fresh. The benefit of that is that each test gets its own metadata state (stored in `scratchFetch.js`, which is also pulled in by `FetchTool.js`). The downside is that each test needs to do its own non-global `require` (or `import`) each time.

I'm basically following the pattern shown here: https://jest-bot.github.io/jest/docs/jest-object.html#jestresetmodules

delasare: Cool. Thanks for the explanation!

PR #419, opened 2023-09-29
Control scratchFetch with a query param

cwillisf:
### Proposed Changes

Don't add metadata in `scratchFetch` unless the location includes `scratchMetadata=1`

### Reason for Changes

This should let us test the metadata changes in production without the risk of breaking the site for everyone.

### Test Coverage

Tested locally using `npm link` and logpoints in Chrome.

### DIFF
--- a/src/scratchFetch.js
+++ b/src/scratchFetch.js
@@ -23,6 +23,19 @@ const metadata = new crossFetch.Headers();
  * @returns {boolean} true if `metadata` has contents, or false if it is empty.
  */
 const hasMetadata = () => {
+    /* global self */
+    const searchParams = (
+        typeof self !== 'undefined' &&
+        self &&
+        self.location &&
+        self.location.search &&
+        self.location.search.split(/[?&]/)
+    ) || [];
+    if (!searchParams.includes('scratchMetadata=1')) {
+        // for now, disable this feature unless scratchMetadata=1
+        // TODO: remove this check once we're sure the feature works correctly in production
+        return false;
+    }
     for (const _ of metadata) {
         return true;
     }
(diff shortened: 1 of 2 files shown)


src/scratchFetch.js
```diff
@@ -23,6 +23,19 @@ const metadata = new crossFetch.Headers();
  * @returns {boolean} true if `metadata` has contents, or false if it is empty.
  */
 const hasMetadata = () => {
+    /* global self */
+    const searchParams = (
+        typeof self !== 'undefined' &&
+        self &&
+        self.location &&
+        self.location.search &&
+        self.location.search.split(/[?&]/)
```
cwillisf, line 32: Splitting `?abc=1&def=2` this way will result in `['', 'abc=1', 'def=2']`. That empty string won't interfere with the `includes` check below, so I didn't worry about ignoring the `?` or dropping the first element of the array.

colbygk: While I want to avoid the perfect being the enemy of the good, I worry each time we include query parameters as part of any interactions for HTTP related transactions.

While free form query parameters are good for speedy development/ease of implementation, they also contribute to ease of cache busting and making it more complex for us to work at scale with vetting/allowlist/blocklisting query parameters as far away from the core application servers as possible, i.e. block things that look obviously wrong as close to the edge of our product as possible (i.e. Fastly right now). There's the trade off here of how we use/create such tooling that it can create boiler plate handling code/dictionaries both for the edge and for the individual localized development environment, if we don't use query parameters we still need boilerplate that would cover headers that provide this and there may indeed be issues that make the use of query parameters required... If nothing else, we should start codifying these parameters so that we can do some checking/filtering at the edge using some automated tooling to configure it...

colbygk (approved):
LGTM

PR #421, opened 2023-10-05
Fix scratch-storage@2.3.x for browser builds

cwillisf:
### Resolves

- Likely resolves ENG-107

### Proposed Changes

Update the `browser` field in `package.json` to point to `dist/web/scratch-storage.js` instead of `src/index.js`

### Reason for Changes

Browsers don't understand `webpack` loaders, so strictly speaking, this field has been incorrect since we first implemented default assets. We didn't notice because, within Scratch repositories, we always build with `webpack`. Removing the inline loader syntax in #412 made it so that dependents need to specify loader rules, and thus broke `scratch-gui` and others. Telling dependents to use the build output makes it so the loader is applied as part of the `scratch-storage` build process and doesn't need to be configured in every dependent. Also, FLUFFY FISHSTICKS this took SO much longer than it should have. It's exceptionally difficult to fix this at the `scratch-gui` level, which is what I was beating my head against for a few days. I'm very happysad that it's effectively a one-liner in `scratch-storage`.

### Test Coverage

I enhanced the `load-default-assets` test to verify the exact size of each default asset. That way, we can be sure that the file contents are accurate instead of, for example, a filename or a data URI string. Basically, it now tests that the loader is doing its job and configured correctly. To do this, it needs to use the build output instead of the raw source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/LLK/scratch-storage.git"
   },
   "main": "./dist/node/scratch-storage.js",
-  "browser": "./src/index.js",
+  "browser": "./dist/web/scratch-storage.js",
   "scripts": {
     "build": "webpack --progress --colors --bail",
     "coverage": "tap ./test/{unit,integration}/*.js --coverage --coverage-report=lcov",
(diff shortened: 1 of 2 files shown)


AlexisGoodfellow (approved):
YOLO!

But for real, this looks good. Reasoning is very sound. Thanks for the context!

cwillisf:
The "good" news is: if this doesn't fix the problem, `scratch-gui` will still be broken and we're no worse off than yesterday. Hooray!(?)

PR #428, opened 2023-12-21
Update semantic release config

cwillisf:
### Resolves

Resolves ENA-291

### Proposed Changes

Update the versions and configuration for `semantic-release` and friends to match our current standards.


skripted-io (approved):
👊

PR #649, opened 2024-10-10
Upgrade webpack and convert to TS

georgyangelov:
@cwillisf I am weak. I went down the rabbit hole (again). I found the rabbit... actually, I found at least 4-5 hostile rabbits there. 😅 

So I didn't want to have ScratchStorage as `any` so I decided to add the types here. But scratch-storage uses webpack 4 so I thought "how hard can it be to upgrade it". Which lead to web workers packaging which changed in 5 (it's a separate chunk now, not inlined into the main bundle). Which is a breaking change for the lib so => new major version.

Had to also fight jest for the workers, and some specifics around how TS generates `module.exports.default = ...` and not `module.exports = ...` so I changed some exports to be non-default (don't really like default ones https://blog.gangelov.net/posts/why-named-exports/ ).

Anyway, I have this PR with what I believe are no (or minimal) behavior changes, but we have typings which can be used by `scratch-web` for the NGP.

I planned to change a bit the saving logic so that it's not directly going to the webHelper, but allows for configurability, so that we can have, for example, the IndexedDB store I hacked around locally (don't ask - `storage.webHelper = { ... }` 😅).
Anyway - I'll see if this is necessary for now. And that would be a separate PR. I'm thinking maybe something similar to express middlewares - where each store can either proceed to the next one or short-circuit. Does that sound alright on a higher-level?

Also I realize I haven't asked you whether you're ok with having TS in here (or the other changes), so if you have any concerns - I'm definitely open to a discussion.


georgyangelov:
Okay, decided that it's not worth doing any more changes now. With that this set of changes only stays as "upgrade webpack and add typings", so @cwillisf if you prefer to give this more time to mature, I can add some temporary types in `scratch-gui` or `scratch-web` directly and not have to rely on this PR at all. Would that be better?

See https://github.com/scratchfoundation/scratch-editor/pull/32 and https://github.com/scratchfoundation/scratch-platform/pull/60 for how it's used in the current state.

jest.config.js
```diff
@@ -1,5 +1,37 @@
...
     transform: {
+        ...createDefaultEsmPreset({
+            tsconfig: 'tsconfig.test.json',
+
+            // The webpack 5 way to include web workers is to use
+            // `new Worker(new URL('./worker.js', import.meta.url));`.
+            // See https://webpack.js.org/guides/web-workers/
+            // However, the `import.meta.url` is ESM-only and Jest's support for ESM is
+            // still experimental. So, we need to mock it instead (or use experimental
+            // jest & node features).
```
cwillisf, line 20: File this under "justifications for future migration to Playwright"?

package.json
```diff
@@ -11,8 +11,9 @@
   },
   "main": "./dist/node/scratch-storage.js",
   "browser": "./dist/web/scratch-storage.js",
+  "types": "./dist/types/index.d.ts",
   "scripts": {
-    "build": "webpack --progress --colors --bail",
+    "build": "rm -rf dist && webpack",
```
cwillisf:
```suggestion
    "build": "rimraf dist && webpack",
```
for Windows compatibility

src/Asset.ts
```diff
@@ -0,0 +1,86 @@
+import md5 from 'js-md5';
+import {memoizedToString, _TextEncoder, _TextDecoder} from './memoizedToString';
+import {AssetType} from './AssetType';
+import {Buffer} from 'buffer';
+import {DataFormat} from './DataFormat';
+
+// TODO: The comments in this file indicate that the asset id is a string only, but
+// the types in BuiltinHelper and the default project builder in scratch-gui
+// allow for it to be a number as well.
```
cwillisf, line 8: This one's a good advertisement for TS -- or at least good JSDoc hints.

I think `string` is the "correct" option (for now: a string that looks like an integer; later: probably a string that looks like a UUID). I think this is a reasonable approach given our current reality. Thanks for identifying where the numbers are coming from!

src/DataFormat.ts
```diff
@@ -11,6 +11,6 @@ const DataFormat = {
     SB3: 'sb3',
     SVG: 'svg',
     WAV: 'wav'
-};
+} as const;
 
-module.exports = DataFormat;
+export type DataFormat = typeof DataFormat[keyof typeof DataFormat];
```
cwillisf, line 16: What does `typeof DataFormat[keyof typeof DataFormat]` do? Is that a common TS idiom?

src/FetchWorkerTool.ts
```diff
@@ -33,22 +45,26 @@ class PrivateFetchWorkerTool {
...
+                );
 
                 worker.addEventListener('message', ({data}) => {
                     if (data.support) {
                         this._workerSupport = data.support;
                         return;
                     }
                     for (const message of data) {
-                        if (this.jobs[message.id]) {
+                        const job = this.jobs[message.id];
```
cwillisf, line 62: 😎

src/ScratchStorage.ts
```diff
@@ -182,15 +200,14 @@ class ScratchStorage {
      *   If the promise is rejected, there was an error on at least one asset source. HTTP 404 does not count as an
      *   error here, but (for example) HTTP 403 does.
      */
-    load (assetType, assetId, dataFormat) {
-        /** @type {Helper[]} */
+    load (assetType: AssetType, assetId: AssetId, dataFormat: DataFormat): Promise<_Asset | null> {
         const helpers = this._helpers.map(x => x.helper);
-        const errors = [];
+        const errors: unknown[] = [];
```
cwillisf, line 210: I'm still a little embarrassed that I didn't use `Error` throughout :/

src/memoizedToString.js
```diff
@@ -0,0 +1,75 @@
...
+    // eslint-disable-next-line global-require
+    const encoding = require('fastestsmallesttextencoderdecoder');
+    _TextDecoder = encoding.TextDecoder;
+    _TextEncoder = encoding.TextEncoder;
+} else {
+    _TextDecoder = TextDecoder;
+    _TextEncoder = TextEncoder;
+}
+
+const memoizedToString = (function () {
```
cwillisf: Thanks for pulling this out of `Asset`!

cwillisf (approved):
Awesome!

"I am weak." :laughing:
I guess your weakness leads to strength!

Thanks for doing this. It looks great! My only request is to change `rm -rf` to `rimraf` (and add that as a dev dep) to help developers on Windows.

cwillisf:
> I planned to change a bit the saving logic so that it's not directly going to the webHelper, but allows for configurability

> I'm thinking maybe something similar to express middlewares - where each store can either proceed to the next one or short-circuit. Does that sound alright on a higher-level?

That all sounds good to me! The Express-style middleware pattern seems like a good choice, both for the way it works and for familiarity among JS / TS devs.

jest.config.js
georgyangelov: Probably 🤷  Depends on what level we're testing on - E2E through the UI - definitely. Unit-level JS logic - jest still seems the best option

package.json
georgyangelov: Ah, I knew there had to be some reason to use an npm module 😄

src/Asset.ts
georgyangelov: I believe the only **value** that I saw setting it to a number is this in defaultProject in GUI:

```
{
        id: 0,
        assetType: 'Project',
        dataFormat: 'JSON',
        data: JSON.stringify(projectJson)
    }
```

Should I change that to `"0"` and make the AssetId type be just `string`?

src/DataFormat.ts
georgyangelov: Ah, yes, it's basically saying "give me the type of all possible values that DataFormat can have". So that would be the type `'"jpg" | "json" | "mp3" | "png" | "sb2" | "sb3" | "svg" | "wav"`. We can also say `string`, but since they are an enumeration I wanted to be more specific just in case somewhere this pays off.

In  `typeof DataFormat[keyof typeof DataFormat]`:
- `typeof DataFormat` is `{JPG: 'jpg', JSON: 'json', ...}`
- `keyof {JPG: 'jpg', JSON: 'json', ...}` is `"JPG" | "JSON" | ...` - basically the type of all possible keys as a union
- `{JPG: 'jpg', JSON: 'json'}['JPG']` is `'jpg'` - like getting the value of a property, but for types
- And you can have unions as the indexing: `{JPG: 'jpg', JSON: 'json', PNG: 'png'}['JPG' | 'PNG']` gets you `'jpg' | 'png'` (basically following all versions of the union)

src/FetchWorkerTool.ts
georgyangelov: Had to please TS here since it's an access and it doesn't know that it's non-null later 😄

src/ScratchStorage.ts
georgyangelov: I didn't check, to be honest, but having promise rejections be `unknown` in TS is standard practice since any rejection up the chain can be an arbitrary object and this can even be libraries. So even if we use Error all around our code, it still is hard to make sure it's an Error in every possible case.

src/memoizedToString.js
georgyangelov: I didn't want to have the `require` above in TS and thought it safer to leave in JS for now :D

georgyangelov:
Thank you for reviewing and being open to bigger changes! As a next step, I'll finish up the scratch-gui changes and make 100% sure all tests in there pass as well.

Do you think I should test in some other way before merging this? Some other project that uses scratch-storage?
Or something to document somewhere (readme doesn't seem to need changes, but maybe document something?)

georgyangelov:
I think I'm done with this, I corrected a couple more types and fixed the node bundle that I broke previously. https://github.com/scratchfoundation/scratch-storage/pull/649/commits/ff5b428e7152b97671262da1d7ac08f3abfe7259 this is a breaking change as well, but I guess it should be fine to not update the non-monorepo scratch-gui?

I can also backport the change, what do you think?

PR #660, opened 2024-10-23
fix: fix web worker usage in scratch-vm integration tests

georgyangelov:
See https://github.com/lquixada/cross-fetch/issues/78 for why it was necessary to update cross-fetch.
Also needed to expose getMetadata as with webpack 5 it seems to get eliminated (as we're building with NODE_ENV=production now due to some auto-optimizations with webpack 5)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -27,10 +27,11 @@ if (!process.env.CI) {
 const webConfig = baseConfig.clone()
     .merge({
         output: {
-            library: 'ScratchStorage',
-            libraryTarget: 'umd',
+            library: {
+                name: 'ScratchStorage',
+                type: 'umd2'
+            },
             path: path.resolve(__dirname, 'dist', 'web'),
-            filename: '[name].js',
             clean: false
         }
     });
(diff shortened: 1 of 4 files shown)


georgyangelov:
@cwillisf since you're OOO and this is a fix for the 3.x version not used by anything at the moment, I'll merge it with @MiroslavDionisiev's approval. 🙏

webpack.config.js
```diff
@@ -63,6 +63,7 @@ const nodeConfig = baseConfig.clone()
         },
         output: {
             libraryTarget: 'commonjs2',
```
MiroslavDionisiev: Webpack's documentation suggests moving to `library.type` instead. In `scratch-webpack-configuration` `library.type` is set to `umd2`. Wouldn't that be conflicting? Probably should do:
```suggestion
            library: {
                type: 'commonjs'
            }
```
Also some of the other fields are already part of the shared configuration

MiroslavDionisiev (approved):

webpack.config.js
georgyangelov: Thank you, updated the config

PR #716, opened 2025-01-06
fix: error calling fetch

MiroslavDionisiev:
### Resolves

Addresses error `'fetch' called on an object that does not implement interface Window.` when using `cross-fetch`


georgyangelov (approved):

cwillisf (approved):
I tested this locally and can verify that it fixes the error.
Thanks!

PR #885, opened 2026-03-16
feat: make WebHelper support async UrlFunctions

georgyangelov:
Would be useful for apps that need to resolve some sort of short-lived token asynchronously before they can construct the request object.

Also refactored the code structure to use async/await instead of chaining promises

### DIFF
--- a/src/WebHelper.ts
+++ b/src/WebHelper.ts
@@ -144,12 +144,12 @@ export default class WebHelper extends Helper {
             const reqConfigFunction = store && store.get;
 
             if (reqConfigFunction) {
-                const reqConfig = await ensureRequestConfig(reqConfigFunction(asset));
-                if (!reqConfig) {
-                    continue;
-                }
-
                 try {
+                    const reqConfig = await ensureRequestConfig(reqConfigFunction(asset));
+                    if (!reqConfig) {
+                        continue;
+                    }
+
                     const body = await tool.get(reqConfig);
                     if (body) {
                         asset.setData(body, dataFormat);
--- a/src/WebHelper.ts
+++ b/src/WebHelper.ts
@@ -187,50 +187,60 @@ export default class WebHelper extends Helper {
         // If we have an asset id, we should update, otherwise create to get an id
         const create = assetId === '' || assetId === null || typeof assetId === 'undefined';
 
-        // Use the first store with the appropriate asset type and url function
-        const store = this.stores.filter(s =>
+        const candidateStores = this.stores.filter(s =>
             // Only use stores for the incoming asset type
             s.types.indexOf(assetType.name) !== -1 && (
                 // Only use stores that have a create function if this is a create request
                 // or an update function if this is an update request
                 (create && s.create) || s.update
             )
-        )[0];
+        );
 
         const method = create ? 'post' : 'put';
 
-        if (!store) return Promise.reject(new Error('No appropriate stores'));
+        if (candidateStores.length === 0) {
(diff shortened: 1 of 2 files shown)


src/WebHelper.ts
```diff
@@ -120,40 +140,33 @@ export default class WebHelper extends Helper {
...
             const reqConfigFunction = store && store.get;
 
             if (reqConfigFunction) {
-                const reqConfig = ensureRequestConfig(reqConfigFunction(asset));
-                if (reqConfig === false) {
-                    return tryNextSource();
+                const reqConfig = await ensureRequestConfig(reqConfigFunction(asset));
+                if (!reqConfig) {
+                    continue;
                 }
```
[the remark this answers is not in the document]
georgyangelov: Good point, updated

src/WebHelper.ts
```diff
@@ -193,30 +206,31 @@ export default class WebHelper extends Helper {
...
         }
 
-        const reqConfig = ensureRequestConfig(
+        const reqConfig = await ensureRequestConfig(
             // The non-nullability of this gets checked above while looking up the store.
             // Making TS understand that is going to require code refactoring which we currently don't
             // feel safe to do.
             create ? store.create!(asset) : store.update!(asset)
         );
         const reqBodyConfig = Object.assign({body: data, method}, reqConfig);
```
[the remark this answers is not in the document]
georgyangelov: Nice, fixed

src/WebHelper.ts
```diff
@@ -174,49 +187,60 @@ export default class WebHelper extends Helper {
...
+            // xhr makes it difficult to both send FormData and
+            // automatically parse a JSON response. So try to parse
+            // everything as JSON.
+            if (typeof body === 'string') {
+                try {
+                    body = JSON.parse(body);
+                } catch (parseError) { // eslint-disable-line @typescript-eslint/no-unused-vars
+                    // If it's not parseable, then we can't add the id even
+                    // if we want to, so stop here
+                    return body;
```
KManolov3, line 235: If there are multiple candidates, wouldn't we want to continue here and try one of the others? Or are we okay with returning this to preserve current behaviour?

KManolov3 (approved):
Looks good to me!

src/WebHelper.ts
georgyangelov: I was trying to preserve current behavior, not sure if this happens actually

src/WebHelper.ts
```diff
@@ -120,40 +140,33 @@ export default class WebHelper extends Helper {
...
 
-        let storeIndex = 0;
-        const tryNextSource = (err?: unknown): Promise<Asset | null> => {
-            if (err) {
-                errors.push(err);
-            }
-            const store = stores[storeIndex++];
-
-            /** @type {UrlFunction} */
+        for (const store of stores) {
```
cwillisf, line 143: THANK YOU

I've been meaning to fix up this promise chain for ages 😵‍💫

src/WebHelper.ts
cwillisf: Thinking about all the possible theoretical situations, I think trying the next candidate is often the right answer, but not always. Thinking about the situations where this code is actually used, I believe it only ever has one candidate, so I don't think it matters ;)

I think my vote is to not change current behavior right now, but I'm OK with trying the next candidate if we both a) find a use case and b) commit to careful QA on the change.

cwillisf (approved):

